PR #14, opened 2017-08-18
PoC functionalities

### LINKED ISSUE #8: Write Readme on how to install and start the relay server

### LINKED ISSUE #9: Load json abi to talk to contracts.
see TODO trustlines.py
load the generated abi from populus

### LINKED ISSUE #10: Add missing parts of REST api
add logic to resources.py and define paths in app.py
use the proxies or the graph to process the query. Most functionality should be there already.

dakingha69:
This closes #8, closes #9 and closes #10


berndbohmeier:
Looks good. Pretty much like I would have liked to have it.

I only dont like the fact that the contract.json is now fixed in here, because if it changes in the other repo we need to manually update it. On the other hand that means that changes not immediately break things. We should probably think of a way of doing this also for the abi in the client lib and if we want to keep it that way or change it.

PR #55, opened 2018-02-26
Add basic smoketest

### LINKED ISSUE #37: Create a smoketest
Simple smoketest that starts the relay server and does a simple query to check if it running

berndbohmeier:
Closes #37


berndbohmeier:
Still need to find out, why it fails, works locally

berndbohmeier:
totally outdated, but I still think we should have a smoke test.

PR #74, opened 2018-04-17
Fix/graph test

cducrest:
I did a rebase and corrected the tests.
Should have done a "rebase -i" to make it cleaner but read about it later.

### DIFF
--- a/relay/network_graph/dijkstra_weighted.py
+++ b/relay/network_graph/dijkstra_weighted.py
@@ -79,6 +79,7 @@ def find_path(G, source, target, get_fee, value, max_hops=None, max_fees=None):
     push = heappush
     pop = heappop
     dist = {}  # dictionary of final distances
+    dist_hop = {}  # dictionary of final distances in terms of hops
     seen = {source: value}
     c = count()
     fringe = []  # use heapq with (distance,label) tuples
@@ -88,6 +89,7 @@ def find_path(G, source, target, get_fee, value, max_hops=None, max_fees=None):
         if v in dist:
             continue  # already searched this node.
         dist[v] = d
+        dist_hop[v] = n
         if v == target:
             break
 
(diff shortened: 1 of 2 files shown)


relay/network_graph/dijkstra_weighted.py
```diff
@@ -103,7 +105,7 @@ def find_path(G, source, target, get_fee, value, max_hops=None, max_fees=None):
                 if n + 1 > max_hops:
                     continue
             if u in dist:
-                if vu_dist < dist[u]:
+                if n+1 < dist_hop[u]:  # will have to be updated in case we do a merged cost for hops and fees
```
berndbohmeier: We discussed that we want to use the same logic like nodes are popped.

tests/unit/test_graph.py
```diff
@@ -168,6 +185,15 @@ def test_no_direction(community_with_trustlines):
     assert path == []
 
 
+def test_valid_path_raise_ValueError(complex_community_with_trustlines_and_fees):
```
berndbohmeier: Did you mean test_valid_path_raises_no_value_error? 
Also please dont use CamelCase in functions in python. 
If you write a test for a previous bug you should write a comment with the issue number or describe the bug here. Otherwise this test might get deleted in the future because no one know why you test something random that should work.

relay/network_graph/dijkstra_weighted.py
cducrest: I put a better logic for the test, removed the test on dist_hop for more visibility and because it logically cannot fail since we are using integer that cannot be negative and not costs functions.

tests/unit/test_graph.py
cducrest: I did mean "test_valid_path_raise_ValueError" , it tests whether a valid path raises a ValueError and does not assert anything on the result (but I changed it if you like it). It was CamelCase because it refers to the name of the error.
Also, I wrote the test to show that the bug was corrected, and do not mind if it gets deleted in the future.

berndbohmeier: Well I would mind. It is now a regression test and should stay.

PR #76, opened 2018-04-18
Wrapped ETH balance

### LINKED ISSUE #67: Endpoint for wrapped ETH balance
We need an endpoint for retrieving the balance of wrapped ETH of an user. I am not sure though how the endpoint should be named. I think something like `users/:userAddress/balances/:ethTokenAddress` could be the most consistent regarding our existing endpoints. Or maybe even use `users/:userAddress/balances` and return an object? @berndbohmeier what do you think?

dakingha69:
Closes #67

- adds `/tokens` endpoint
- adds `/tokens/:tokenAddress/users/:userAddress/balance` endpoint

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: python
 python:
   - "3.6"
 install:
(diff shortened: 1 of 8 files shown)


dakingha69:
move changes to PR #78.

PR #79, opened 2018-04-24
Order details endpoint with filled, cancelled and unavailable amounts

### LINKED ISSUE #72: Endpoint for order details
Add endpoint to retrieve a single order by its hash `GET /exchange/order/:orderHash`. Maybe also add parameters to optionally retrieve filled, canceled and unavailable amount of specific order. Something like `/exchange/order/:orderHash?filled=true&canceled=true&unavailable=true` whereas unavailable is the sum of filled and canceled amount.

dakingha69:
Closes #72

### DIFF
--- a/relay/api/exchange/resources.py
+++ b/relay/api/exchange/resources.py
@@ -55,7 +55,36 @@ class OrderBook(Resource):
         }
 
 
+class OrderDetail(Resource):
+
+    def __init__(self, trustlines: TrustlinesRelay) -> None:
+        self.trustlines = trustlines
+
+    args = {
+        'includeFilled': webfields.Boolean(required=False, missing=False),
+        'includeCancelled': webfields.Boolean(required=False, missing=False),
+        'includeUnavailable': webfields.Boolean(required=False, missing=False)
+    }
+
+    @use_args(args)
+    def get(self, args, order_hash: str):
+        order_orm = self.trustlines.orderbook.get_order_by_hash(bytes.fromhex(order_hash[2:]))
+
+        if order_orm is None:
+            abort(422, message='Order does not exist')
+
+        order = order_as_dict(order_orm)
+        if (args['includeFilled']):
+            order['filledTakerAmount'] = self.trustlines.orderbook.get_filled_amount(order_orm)
+        if (args['includeCancelled']):
+            order['cancelledTakerAmount'] = self.trustlines.orderbook.get_cancelled_amount(order_orm)
+        if (args['includeCancelled']):
+            order['unavailableTakerAmount'] = self.trustlines.orderbook.get_unavailable_amount(order_orm)
+        return order
+
+
 class OrderSubmission(Resource):
+
     def __init__(self, trustlines: TrustlinesRelay) -> None:
         self.trustlines = trustlines
 
--- a/relay/exchange/orderbook.py
+++ b/relay/exchange/orderbook.py
@@ -83,6 +83,10 @@ class OrderBook(object):
         if self._db is not None:
             return self._db.order_filled(orderhash, filled_maker_amount, filled_taker_amount)
 
+    def get_order_by_hash(self, order_hash: bytes):
+        if self._db is not None:
+            return self._db.get_order_by_hash(order_hash)
+
 
 class OrderBookGreenlet(OrderBook):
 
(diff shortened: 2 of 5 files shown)


relay/api/exchange/resources.py
```diff
@@ -55,7 +55,36 @@ def get(self, args):
         }
 
 
+class OrderDetail(Resource):
+
+    def __init__(self, trustlines: TrustlinesRelay) -> None:
+        self.trustlines = trustlines
+
+    args = {
+        'includeFilled': webfields.Boolean(required=False, missing=False),
```
berndbohmeier: Why do we need those options? I would just include them always?

relay/api/exchange/resources.py
```diff
@@ -55,7 +55,36 @@ def get(self, args):
...
+
+    @use_args(args)
+    def get(self, args, order_hash: str):
+        order_orm = self.trustlines.orderbook.get_order_by_hash(bytes.fromhex(order_hash[2:]))
+
+        if order_orm is None:
+            abort(422, message='Order does not exist')
+
+        order = order_as_dict(order_orm)
+        if (args['includeFilled']):
```
berndbohmeier: the available amount should be in there already https://github.com/trustlines-network/relay/blob/33bde27ea4eec43c4d89c1473d3fb526d753b267/relay/exchange/exchange_db.py#L69. 
Or is there something wrong with the exsiting code?

relay/api/exchange/resources.py
```diff
@@ -55,7 +55,36 @@ def get(self, args):
...
+            order['filledTakerAmount'] = self.trustlines.orderbook.get_filled_amount(order_orm)
+        if (args['includeCancelled']):
+            order['cancelledTakerAmount'] = self.trustlines.orderbook.get_cancelled_amount(order_orm)
```
berndbohmeier: Those are calls to the ethereum node and are to slow. We should query that from our database. I thought it is there already https://github.com/trustlines-network/relay/blob/33bde27ea4eec43c4d89c1473d3fb526d753b267/relay/exchange/exchange_db.py#L69. Otherwise please add there what is missing.

berndbohmeier: Ah okay, atm you only get the unavailable amounts. If you need them separate, you should add that place https://github.com/trustlines-network/relay/blob/33bde27ea4eec43c4d89c1473d3fb526d753b267/relay/exchange/exchange_db.py#L46

relay/api/exchange/resources.py
```diff
@@ -55,7 +61,20 @@ def get(self, args):
         }
 
 
+class OrderDetail(Resource):
+
+    def __init__(self, trustlines: TrustlinesRelay) -> None:
+        self.trustlines = trustlines
+
+    def get(self, order_hash: str):
+        order_orm = self.trustlines.orderbook.get_order_by_hash(bytes.fromhex(order_hash[2:]))
```
berndbohmeier: it is not an order_orm but just an order

relay/exchange/orderbook.py
```diff
@@ -83,6 +83,14 @@ def order_filled(self, orderhash: bytes, filled_maker_amount: int, filled_taker_
         if self._db is not None:
             return self._db.order_filled(orderhash, filled_maker_amount, filled_taker_amount)
 
+    def order_cancelled(self, orderhash: bytes, cancelled_maker_amount: int, cancelled_taker_amount: int):
```
berndbohmeier: Please add return type and change it to not return the called function, because it always returns None. Same in order_filled, was badly written by me I guess.

relay/api/exchange/resources.py
berndbohmeier: This is just and order, not an order_orm. Just change the variable name.

relay/api/exchange/resources.py
```diff
@@ -55,7 +61,20 @@ def get(self, args):
...
+        if order_orm is None:
+            abort(422, message='Order does not exist')
```
berndbohmeier: I would say a simple 404 error code fits better than a 422

relay/api/exchange/resources.py
berndbohmeier: You have to validate that the provided string is a valid hash. Right now just trying for exapme 1234 for the hash results in an internal server error: ValueError: non-hexadecimal number found in fromhex() arg at position 0

berndbohmeier:
LGTM.

PR #84, opened 2018-05-08
Added function to find 'triangulation path', sending money to one's self

cducrest:
See issue: https://github.com/trustlines-network/relay/issues/32

### DIFF
--- a/relay/network_graph/dijkstra_weighted.py
+++ b/relay/network_graph/dijkstra_weighted.py
@@ -56,21 +56,22 @@ def find_path(G, source, target, get_fee, value, max_hops=None, max_fees=None, i
 
 def find_path_triangulation(G, source, target_reduce, target_increase, get_fee, value, max_hops=None, max_fees=None):
     """
-    target_reduce is the node we want to reduce our credit with (for value > 0)
-    target_increase is the node which will result with an increased debt (for value > 0)
-    source will owe less to target_reduce and target_increase will owe more to source (for value > 0)
+    target_reduce is the node we want to reduce our credit with
+    target_increase is the node which will result with an increased debt
+    source will owe less to target_reduce and target_increase will owe more to source
     must return a path and the total fee of that path
     to be called in the right order with source as source and target as target, contrary to find_path
-
-    value may be negative, in which case
-    source will owe more to target_reduce and target_increase will owe less to source
+    value must be >0
     """
     def get_fee_wrapper(b, a, value):
         # used to get the data from the graph in the right order and query the fees
         if b < a:
-            return get_fee(b, a, G[b][a], value)
+            output = get_fee(b, a, G[b][a], value)
         else:
-            return get_fee(b, a, G[a][b], value)
+            output = get_fee(b, a, G[a][b], value)
+        if output is None:
+            raise nx.NetworkXNoPath("node %s not reachable from %s" % (a, b))
+        return output
 
     def verify_balance_greater_than_value(a, b, value):
         # used to verify that we reduce the amount source owes to target_reduce and do not misuse the function
@@ -79,16 +80,12 @@ def find_path_triangulation(G, source, target_reduce, target_increase, get_fee, 
         else:
             return G[b][a]['balance_ab'] >= value
 
-    def output(final_fee, intermediary_path):
-        if max_fees is not None and final_fee > max_fees:
-            raise nx.NetworkXNoPath(
-                "operation impossible due to fees: %d with max_fees: %d" % (final_fee, max_fees))
-        return (final_fee, list(reversed([source] + intermediary_path + [source])))
-
+    # verification that the funtion is used properly
     if value <= 0:
         raise ValueError('This value cannot be handled yet : %d' % value)
     elif not verify_balance_greater_than_value(source, target_reduce, value):
-        raise ValueError('The value is greater than the amount source owes target_reduce : %d' % value)
+        raise nx.NetworkXNoPath(
+            "The balance of target_reduce is lower than value %d" % value)
 
     G_succ = G.succ if G.is_directed() else G.adj
     neighbors = [x[0] for x in G_succ[source].items()]
@@ -99,33 +96,22 @@ def find_path_triangulation(G, source, target_reduce, target_increase, get_fee, 
     if target_increase not in neighbors:
         raise nx.NetworkXNoPath(
             "node %s not a neighbor of %s" % (target_increase, source))
+    if target_increase == target_reduce:
+        raise ValueError("target_increase is equal target_reduce: %s" % target_increase)
 
-    if value >= 0:
-        first_fee = get_fee_wrapper(source, target_reduce, value)
+    first_fee = get_fee_wrapper(source, target_reduce, value)
 
-        if max_hops is not None:
-            max_hops -= 2
+    if max_hops is not None:
+        max_hops -= 2
 
-        intermediary_fee, intermediary_path = find_path(G, target_reduce, target_increase,
-                                                        get_fee, value+first_fee, max_hops, max_fees, source)
+    intermediary_fee, intermediary_path = find_path(G, target_reduce, target_increase,
+                                                    get_fee, value+first_fee, max_hops, max_fees, source)
 
-        last_fee = get_fee_wrapper(target_increase, source, value+intermediary_fee+first_fee)
+    last_fee = get_fee_wrapper(target_increase, source, value+intermediary_fee+first_fee)
 
-        final_fee = last_fee + intermediary_fee + first_fee
-        return output(final_fee, intermediary_path)
+    final_fee = last_fee + intermediary_fee + first_fee
 
-    else:
-        value = -value
-
-        first_fee = get_fee_wrapper(source, target_increase, value)
-
-        if max_hops is not None:
-            max_hops -= 2
-
-        intermediary_fee, intermediary_path = find_path(G, target_increase, target_reduce, get_fee,
-                                                        value+first_fee, max_hops, max_fees, source)
-
-        last_fee = get_fee_wrapper(target_reduce, source, value + intermediary_fee + first_fee)
-
-        final_fee = last_fee + intermediary_fee + first_fee
-        return output(final_fee, intermediary_path)
+    if max_fees is not None and final_fee > max_fees:
+        raise nx.NetworkXNoPath(
+            "operation impossible due to fees: %d with max_fees: %d" % (final_fee, max_fees))
+    return (final_fee, list(reversed([source] + intermediary_path + [source])))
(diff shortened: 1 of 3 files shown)


relay/network_graph/dijkstra_weighted.py
```diff
@@ -87,12 +20,14 @@ def find_path(G, source, target, get_fee, value, max_hops=None, max_fees=None):
         (n, d, _, v) = pop(fringe)
         if v in dist:
             continue  # already searched this node.
+        if v == ignore:
```
berndbohmeier: What about calling it ignored_nodes and make it a list, so that also more than one node can be ignored.

relay/network_graph/dijkstra_weighted.py
```diff
@@ -87,12 +20,14 @@ def find_path(G, source, target, get_fee, value, max_hops=None, max_fees=None):
...
+            continue
         dist[v] = d
         if v == target:
             break
 
         for u, e in G_succ[v].items():
-            cost = get_fee(v, u, e, d)
+            cost = get_fee(v, u, e, d)  # fee of transferring from u to v
```
berndbohmeier: Is that actually the fee from u to v? Than we should fix the get_fee function.

relay/network_graph/dijkstra_weighted.py
```diff
@@ -111,7 +46,81 @@ def find_path(G, source, target, get_fee, value, max_hops=None, max_fees=None):
...
+
+    if target_reduce not in neighbors:
+        raise nx.NetworkXNoPath(
+            "node %s not a neighbor of %s" % (target_reduce, source))
+
+    if target_increase not in neighbors:
+        raise nx.NetworkXNoPath(
+            "node %s not a neighbor of %s" % (target_increase, source))
+
+    if value >= 0:
```
berndbohmeier: If you think you can not reduce the code, then I would say those two cases should go to different functions find_path_triangulation_reduce and find_path_triangulation_increase. It only makes sense to combine them if there share something.

relay/network_graph/dijkstra_weighted.py
```diff
@@ -111,7 +46,81 @@ def find_path(G, source, target, get_fee, value, max_hops=None, max_fees=None):
...
+        if source < target_reduce:
```
berndbohmeier: Please fix the get_fee function, that we dont have to do that every time check every time before calling the function.

relay/network_graph/dijkstra_weighted.py
```diff
@@ -111,7 +46,81 @@ def find_path(G, source, target, get_fee, value, max_hops=None, max_fees=None):
...
+                                                        get_fee, value+first_fee, max_hops, max_fees, source)
+
+        if source < target_increase:
+            last_fee = get_fee(target_increase, source, G[source][target_increase], value+intermediary_fee+first_fee)
+        else:
+            last_fee = get_fee(target_increase, source, G[target_increase][source], value+intermediary_fee+first_fee)
+
+        final_fee = last_fee + intermediary_fee + first_fee
+        if max_fees is not None:
+            if last_fee + intermediary_fee + first_fee > max_fees:
```
berndbohmeier: you should use final_fee here

berndbohmeier: And combine the two ifs into one. Its harder to read it like that.

relay/network_graph/dijkstra_weighted.py
berndbohmeier: But I am sure it is possible to do it in one function if you find the similarities in the code and all the a < b if else things in just one line, after fixing the get fee or write your own get_fee function.

relay/network_graph/dijkstra_weighted.py
berndbohmeier: Or you can write a wrapper function that does that.

berndbohmeier: If you feel like this function is not need outside you can also define it as a inner function. https://realpython.com/inner-functions-what-are-they-good-for/

tests/unit/test_graph.py
```diff
@@ -109,6 +126,50 @@ def test_transfer(community_with_trustlines):
...
+def test_triangulation_negative_value_with_cost(complex_community_with_trustlines_and_fees):
+    complex_community_with_trustlines_and_fees.update_balance(A, B, -10000)  # amount B owes A
+    complex_community_with_trustlines_and_fees.update_balance(A, C, 10000)
+    complex_community_with_trustlines_and_fees.update_balance(B, D, -10000)
+    complex_community_with_trustlines_and_fees.update_balance(C, D, 10000)
+    cost, path = complex_community_with_trustlines_and_fees.find_path_triangulation(
+            A, B, C, -1000)
+    assert path == [A, C, D, B, A]
+    assert cost == 44
+
```
berndbohmeier: Please add tests that test that the paths are actual doing what they should. Call the find_path_triangulation function and than execute a transfer with that path and check if the balance was reduces/increased by that amount. Most important, test if you can reduce a balance from a value x to zero.

relay/network_graph/dijkstra_weighted.py
cducrest: Because it just adds complexity while not adding any value, in my opinion.
If I want to ignore a list of node, it is going to be faster to modify the graph directly when I know which node to ignore, otherwise we add nodes to be ignored to the list to visit and actually visit them but do nothing when you visit them. (It would be inefficient)

relay/network_graph/dijkstra_weighted.py
cducrest: Yes, since in the application for a path in reverse, for the first iteration, we have "v = destination", u is one neighbor of destination. We then need the fee of transferring from u to v.

relay/network_graph/dijkstra_weighted.py
berndbohmeier: Okay

relay/network_graph/graph.py
```diff
@@ -353,3 +377,22 @@ def mediated_transfer(self, source, target, value):
             target = source
         assert fees == cost
         return cost
+
+    def triangulation_transfer(self, source, target_reduce, target_increase, value):
+        """simulate triangulation transfer off chain"""
+        cost, path = self.find_path_triangulation(source, target_reduce, target_increase, value)
+        assert path[0] == source
+        assert path[-1] == source
+        path = list(reversed(path))
```
berndbohmeier: from here on it is just copy paste. Please refactored and do not duplicate the code. 
For example have a function transfer_path(value, path)

relay/network_graph/graph.py
```diff
@@ -353,3 +377,22 @@ def mediated_transfer(self, source, target, value):
...
+        fees = 0
+        if value < 0:
```
berndbohmeier: What is this? Isnt this just silently droping an error with a weird outcome?

relay/network_graph/dijkstra_weighted.py
```diff
@@ -111,7 +46,68 @@ def find_path(G, source, target, get_fee, value, max_hops=None, max_fees=None):
...
+
+    first_fee = get_fee_wrapper(source, target_reduce, value)
+
+    if max_hops is not None:
+        max_hops -= 2
+
+    intermediary_fee, intermediary_path = find_path(G, target_reduce, target_increase,
+                                                    get_fee, value+first_fee, max_hops, max_fees, source)
+
+    last_fee = get_fee_wrapper(target_increase, source, value+intermediary_fee+first_fee)
```
berndbohmeier: How do you check, that there is enough capacity left on this trustline to do the transfer there?

relay/network_graph/dijkstra_weighted.py
```diff
@@ -111,7 +46,68 @@ def find_path(G, source, target, get_fee, value, max_hops=None, max_fees=None):
...
+
+    final_fee = last_fee + intermediary_fee + first_fee
+    return output(final_fee, intermediary_path)
```
berndbohmeier: I'd say output now suffers a little bit from that problem that you had with functions, that it is unclear what output is doing from looking at its call. Maybe come up with a more descriptive name or dont use a function there.

relay/network_graph/graph.py
cducrest: No, it was for when we had the functionality of triangulation with a negative value (which cannot work if we have no means of making transfer without fees, as already discussed)

cducrest:
I think it is debatable whether I should raise an error when target_reduce and target_increase are the same, because it is a degenerate cases but it works. Since you pointed out that I need a test for that I added an error, but I am not against removing it.

I think it is weird that nx.NetworkXNoPath will be caught by find_path_triangulation in graph.py and ValueErrors are not caught.
This can be seen in the test, when there is a NetworkXNoPath error, we have path == [], and for ValueError we actually catch the errors.
But this is the way it is done for the regular find_path().

relay/network_graph/dijkstra_weighted.py
```diff
@@ -113,7 +48,71 @@ def find_path(G, source, target, get_fee, value, max_hops=None, max_fees=None):
...
+
+def find_path_triangulation(G, source, target_reduce, target_increase, get_fee, value, max_hops=None, max_fees=None):
+    """
+    target_reduce is the node we want to reduce our credit with (for value > 0)
+    target_increase is the node which will result with an increased debt (for value > 0)
+    source will owe less to target_reduce and target_increase will owe more to source (for value > 0)
+    must return a path and the total fee of that path
+    to be called in the right order with source as source and target as target, contrary to find_path
+
+    value may be negative, in which case
```
berndbohmeier: Only values > 0 are allowed now, we need to document that now.

tests/unit/test_graph.py
```diff
@@ -126,6 +126,114 @@ def test_transfer(community_with_trustlines):
...
+    assert complex_community_with_trustlines_and_fees.get_account_sum(A, B).balance == -9000
+
+
+def test_triangulation_not_enough_capacity(complex_community_with_trustlines_and_fees):
+    """A owes money to B and A wants to reduce that amount with the help of C"""
+    complex_community_with_trustlines_and_fees.update_balance(A, B, -10000)  # amount B owes A
+    complex_community_with_trustlines_and_fees.update_balance(A, C, 10000)
+    complex_community_with_trustlines_and_fees.update_balance(B, D, -10000)
+    complex_community_with_trustlines_and_fees.update_balance(C, D, 10000)
+    with pytest.raises(ValueError):
```
berndbohmeier: This should not raise ValueError but gives a empty path like it is now in all other cases when there is no path.

Or we change it that an exception is thrown when there is no path (would be more pythonic, we could use a default parameter to suppress the exception how it is for example in a dict when looking for a value), but then we want to have an own Exception and it needs to be thrown every time when there is now path, so also in findPath. And this needs to be documented then.

PR #95, opened 2018-05-24
Update API docs

### LINKED ISSUE #83: Update REST API documentation
The documentaion of the endpoints is not up to date.

dakingha69:
Closes #83 
- updates REST API documentation of relay
- does not include exchange endpoints

### DIFF
--- a/docs/RelayAPI.md
+++ b/docs/RelayAPI.md
@@ -7,11 +7,11 @@ The relay component of the trustlines network project provides a REST API.
 ```
 You can use the following base endpoint to talk to the currently deployed KOVAN test setup.
 ```
-https://relay0.testnet.trustlines.network:443/api/v1
+https://relay0.testnet.trustlines.network/api/v1
 ```
 
 ## Response
-- All endpoints return either a number, a JSON object or an array
+- All endpoints return JSON
 - All number values are returned in their smallest unit
 - In case of an error, the response of the relay API will have the following format:
 ```javascript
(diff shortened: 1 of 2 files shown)


docs/RelayAPI.md
```diff
@@ -1,376 +1,504 @@
...
-
-## Base URL
-`/api/v1`
+## Base Endpoint
+```
+<protocol>://<host>:<port>/api/v1
+```
+You can use the following base endpoint to talk to the currently deployed KOVAN test setup.
+```
+https://relay0.testnet.trustlines.network:443/api/v1
```
berndbohmeier: Ich würde den Port rausnehmen, das ist ja der Standard Port.

docs/RelayAPI.md
```diff
@@ -1,376 +1,504 @@
...
+```
 
-## Error message
-In case of an error, the response of the relay API will have the following format:
+## Response
+- All endpoints return either a number, a JSON object or an array
```
berndbohmeier: It's all JSON, a number is also various JSON. I would not write it that complicated. Just write that JSON is returned.

docs/RelayAPI.md
```diff
@@ -1,376 +1,504 @@
...
+### `/networks/:network/events`
+Returns a list of event logs in a currency network.
+#### Request
+```
+GET /networks/:network/events?type=:type&fromBlock=:fromBlock
+```
+#### Parameters
+|Name|Type|Required|Description|
+|-|-|-|-|
+|network|string|YES|Address of currency network|
```
berndbohmeier: I'm am not sure if we should mix them like that. Network is a part of the path and I would not call it parameter. Are they referred to as parameter somewhere?

docs/RelayAPI.md
```diff
@@ -1,376 +1,504 @@
...
+```
+POST /networks/:network/path-info
+```
 ### Parameters
-- `currency` -
-
+|Name|Type|Required|Description|
+|-|-|-|-|
+|network|string|YES|Address of currency network|
+### POST body
```
berndbohmeier: Those are the parameters. 
It should be started somewhere that they are posted as JSON. There should be an example.

docs/RelayAPI.md
```diff
@@ -1,376 +1,504 @@
...
+|user|string|YES|Address of user|
+#### Response
 ```javascript
 {
-    "path": ["0xabc123bb...", "0xeebc3bb...", ...], // addresses of users in path
-    "estimatedGas": "12000000",
-    "fees": "0.12" // fees for calculated path
+	"balance": -1000, // sum over balances of all trustlines user has in currency network
+  "given": 2000, // sum of all creditlines given by user in currency network
+  "received": 3000, // sum of all creditlines received by user in currency network
```
berndbohmeier: Are that actually numbers and not strings? Then this is a big. This should be strings like in the events.

docs/RelayAPI.md
dakingha69: Hmm get your point... But if I look at other REST API docs they also call it parameters. Some are distinguishing between **URL Parameters** and **Data Parameters**. Shall we do that also?

docs/RelayAPI.md
dakingha69: Yes currently numbers are returned. I will open an issue for this.

schmir:
grip is a nice way to preview markdown. https://github.com/joeyespo/grip
I've used it to view your changes and it looks like the inline links won't work.

One thing I already noticed, is that it's not clear what the parameters should be. So, I tried to use the network name as :network parameter, which doesn't work, because it expects an address.

docs/RelayAPI.md
berndbohmeier: Can you point me to an example?

dakingha69: - This is just gist I found where they distinguish between **URL** and **Data** params: https://gist.github.com/iros/3426278
- Infura for example calls both just **Parameters** https://infura.io/docs/api/get/symbolFull
- 0x relayer api calls it **Parameters** and **Payload** https://github.com/0xProject/standard-relayer-api/blob/master/http/v0.md#get-v0token_pairs
- ...

berndbohmeier: Yeah. A distinction would be good. And as I said we need both documentation and example. Like it is in the infura doc.

docs/RelayAPI.md
```diff
@@ -1,376 +1,504 @@
...
-## `TODO` Get user information of currency network
-Returns detailed information of currency network in user context
-
-`GET /networks/:networkAddress/users/:userAddress`
-
-### Response
+#### Parameters
+|Name|Type|Required|Description|
+|-|-|-|-|
+|network|string|YES|Address of currency network|
```
berndbohmeier: We need to document that we use ERC55 checksum addresses

dakingha69:
@schmir Thanks for the tip! The internal links should work now. Also changed the parameter names for more clearness.
@berndbohmeier Also incorporated your change requests.

PR #100, opened 2018-05-29
log errors from load addresses, do not fail silently

schmir:
When starting tl-relay with a non-existent rpc.host configured in config.json,
_load_addresses will raise an exception and the greenlet started in
_start_listen_on_new_addresses will exit.

All of this happens without a single thing being logged:
,----
| 2018-05-29 22:27:40,941[INFO] trustlines: Starting relay server
| 2018-05-29 22:27:40,987[INFO] trustlines: Server is running on ('', 5000)
| 2018-05-29 22:27:40,999[INFO] relay: New network: 0x55bdaAf9f941A5BB3EacC8D876eeFf90b90ddac9
`----

So, we now catch errors, log them and let the greenlet retry after it's sleep
period.

Also log the parameters used for connecting to the RPC host.

### DIFF
--- a/relay/relay.py
+++ b/relay/relay.py
@@ -56,7 +56,7 @@ class TrustlinesRelay:
         self._load_config()
         self._load_contracts()
         self._load_orderbook()
-        logger.info('using RPCProvider with parameters %s' % self.config['rpc'])
+        logger.info('using RPCProvider with parameters {}'.format(self.config['rpc']))
         self._web3 = Web3(
             RPCProvider(
                 self.config['rpc']['host'],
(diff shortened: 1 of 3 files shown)


relay/relay.py
```diff
@@ -56,6 +56,7 @@ def start(self):
         self._load_config()
         self._load_contracts()
         self._load_orderbook()
+        logger.info('using RPCProvider with parameters %s' % self.config['rpc'])
```
berndbohmeier: Lets please use the newer .format for string formatting instead of %.

berndbohmeier: Or do you have arguments for using %?

schmir: It's what I'm used to and it's what is being used in C/JAVA. But I'm also fine with using .format.

PR #103, opened 2018-06-07
fix usage of gevent.joinall, increase timeouts

schmir:
We did run into errors on the production server. Some of the calls to get events
did timeout, and we ran into errors because the code didn't handle that case.

gevent.joinall returns the list of finished greenlets, but most of the code
tried to collect results from all of the greenlets.

### DIFF
--- a/relay/api/resources.py
+++ b/relay/api/resources.py
@@ -213,8 +213,8 @@ class UserEvents(Resource):
                 events.append(gevent.spawn(proxy.get_all_network_events,
                                            user_address=user_address,
                                            from_block=from_block))
-        gevent.joinall(events, timeout=5)
-        flattened_events = list(itertools.chain.from_iterable([event.value for event in events]))
+        finished_greenlets = gevent.joinall(events, raise_error=True, timeout=20)
+        flattened_events = list(itertools.chain.from_iterable([gr.value for gr in finished_greenlets]))
         return UserCurrencyNetworkEventSchema().dump(flattened_events, many=True).data
 
 
--- a/relay/blockchain/currency_network_proxy.py
+++ b/relay/blockchain/currency_network_proxy.py
@@ -147,33 +147,45 @@ class CurrencyNetworkProxy(Proxy):
         self.start_listen_on(TransferEventType, log)
 
     def get_network_events(self, event_name: str, user_address: str=None, from_block: int=0) -> List[BlockchainEvent]:
+        logger.debug("get_network_events: event_name=%s user_address=%s from_block=%s",
+                     event_name,
+                     user_address,
+                     from_block)
         if user_address is None:
-            result = self.get_events(event_name, from_block=from_block)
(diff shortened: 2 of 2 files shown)


relay/blockchain/currency_network_proxy.py
```diff
@@ -147,33 +147,45 @@ def log(log_entry):
         self.start_listen_on(TransferEventType, log)
 
     def get_network_events(self, event_name: str, user_address: str=None, from_block: int=0) -> List[BlockchainEvent]:
+        logger.debug("get_network_events: event_name=%s user_address=%s from_block=%s",
```
berndbohmeier: .format

relay/blockchain/currency_network_proxy.py
```diff
@@ -147,33 +147,45 @@ def log(log_entry):
...
+            spawned_greenlets = [
                 gevent.spawn(self.get_events, event_name, filter1, from_block),
                 gevent.spawn(self.get_events, event_name, filter2, from_block)
             ]
-            gevent.joinall(events, timeout=2)
-            result = list(itertools.chain.from_iterable([event.value for event in events]))
-
-            for event in result:
+            finished_greenlets = gevent.joinall(spawned_greenlets, raise_error=True, timeout=15)
+            if len(finished_greenlets) < len(spawned_greenlets):
```
berndbohmeier: This should not only produce a warning, but should fail. We were clearly not able to perform the requested operation.

relay/blockchain/currency_network_proxy.py
```diff
@@ -147,33 +147,45 @@ def log(log_entry):
...
-        return sorted_events(result)
+        return sorted_events(events)
 
     def get_all_network_events(self, user_address: str = None, from_block: int = 0) -> List[BlockchainEvent]:
         events = [gevent.spawn(self.get_network_events,
                                type,
                                user_address=user_address,
                                from_block=from_block) for type in self.standard_event_types]
-        gevent.joinall(events, timeout=5)
+        gevent.joinall(events, timeout=20)
```
berndbohmeier: Timeout needs to be configurable

relay/api/resources.py
```diff
@@ -213,8 +213,8 @@ def get(self, args, user_address: str):
                 events.append(gevent.spawn(proxy.get_all_network_events,
                                            user_address=user_address,
                                            from_block=from_block))
-        gevent.joinall(events, timeout=5)
-        flattened_events = list(itertools.chain.from_iterable([event.value for event in events]))
+        finished_greenlets = gevent.joinall(events, raise_error=True, timeout=20)
```
berndbohmeier: What does raise_error do? It is not documented in gevent.

relay/blockchain/currency_network_proxy.py
berndbohmeier: Okay makes sense

PR #105, opened 2018-06-11
added endpoint for debt reduce

cducrest:
I don't know why we have cost*2 here : https://github.com/trustlines-network/relay/blob/feature/debt-reduce-endpoint/relay/api/resources.py#L391

I only copied the Path class above, but I do not mind having cost*2 as it is only for checking gas price.
I do repeat myself but I do not know if that matters for api.

### DIFF
--- a/docs/RelayAPI.md
+++ b/docs/RelayAPI.md
@@ -32,6 +32,7 @@ https://relay0.testnet.trustlines.network/api/v1
 - [Total spendable amount of user in currency network](#total-spendable-amount-of-user-in-currency-network)
 - [Spendable amount to other user in currency network](#spendable-amount-to-other-user-in-currency-network)
 - [Transfer path in currency network](#transfer-path-in-currency-network)
+- [Debt reduction path in currency network](#debt-reduction-path-in-currency-network)
 - [All events in currency network](#all-events-in-currency-network)
 - [Events of user in currency network](#events-of-user-in-currency-network)
 ### User context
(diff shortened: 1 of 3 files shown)


docs/RelayAPI.md
```diff
@@ -352,6 +353,54 @@ curl --header "Content-Type: application/json" \
...
+|Name|Type|Required|Description|
+|-|-|-|-|
+|networkAddress|string|YES|Address of currency network|
+#### Data Parameters
+|Name|Type|Required|Description|
+|-|-|-|-|
+|from|string|YES|Address of user who wants to reduce debt|
+|to|string|YES|Address of user who sees debt reduce|
+|via|string|YES|Address of intermediary used to reduce debt|
+|value|string|YES|Amount for debt reduction in smallest unit|
```
berndbohmeier: contradicts with your code, where it is optional.

relay/api/resources.py
```diff
@@ -349,6 +349,59 @@ def post(self, args, network_address: str):
                 'fees': cost}
 
 
+class ReduceDebtPath(Resource):
+
+    def __init__(self, trustlines: TrustlinesRelay) -> None:
+        self.trustlines = trustlines
+
+    args = {
+        'value': fields.Int(required=False, missing=1),
```
berndbohmeier: Contradicts with the documentation. 

Basically do you think it make sense to just ask for a reduce path for not a specific value? Then it should be optional like it is here.

cducrest: Alright, then there is the same contradiction in the endpoint for Path (/networks/:networkAddress/path-info) where in the documentation it says the "value" field is required, but sets it to "1" when missing in the code. I personally do not think it makes sense but kept it similar for consistency.

PR #108, opened 2018-06-14
Use configurable timeouts in api

berndbohmeier:
Add timeout as a config parameter
Add a concurrency joinall helper to make that functionality easy accessable
Check timeout only in outer most call in api
Instead of returning half of the events, return a timeout code.

### DIFF
--- a/config.json
+++ b/config.json
@@ -7,5 +7,6 @@
   },
   "syncInterval": 300,
   "updateNetworksInterval": 120,
+  "eventQueryTimeout": 20,
   "enableEtherFaucet": false
 }
--- a/relay/api/resources.py
+++ b/relay/api/resources.py
@@ -1,4 +1,6 @@
 import tempfile
+import functools
+import logging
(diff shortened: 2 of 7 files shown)


berndbohmeier:
1. I want to remove the gevent dependency in the whole api, because I think it is best to limit the gevent dependency to only some places. So using gevent.with_timeout there is not an option for me. Also the timeout parameter ads an abstraction, the user does not need to know in what way we do those concurrent jobs(gevent, async, threads). 
2. The idea was to have the same functionality as gevent.joinall so I added it. I thought the idea is exactly that, it does not matter which of the functions returned first, because they do similar jobs. I can remove it.

schmir:
"I think it is best to limit the gevent dependency to only some places"

That needs some further clarification. What's the reason for that? Do you plan to ditch gevent?

berndbohmeier:
1. We agreed that we do not have to do it that way, but it is okay.
2. Will remove the count argument.

PR #113, opened 2018-06-21
Integrate firebase push notification

### LINKED ISSUE #110: Integration of Firebase

berndbohmeier:
First draft that allows to push events to firebase
Closes #110

Events will be of the form as discussed with jens:
```
{
  "data": {
    "event": "{\"type\": \"Event\", \"timestamp\": 1529591060}"
  },
  "from": "628579622468",
  "priority": "normal",
  "notification": {
    "title": "New Event",
    "body": "Click for more details"
  }
}
```


relay/pushservice/pushservice.py
```diff
@@ -0,0 +1,66 @@
...
+from relay.events import Event
+from relay.api.schemas import UserCurrencyNetworkEventSchema
+from .client_token_db import ClientTokenDB
+
+
+class FirebaseRawPushService:
+    """Sends push notifications to firebase. Sending is done based on raw client tokens"""
+
+    def __init__(self) -> None:
+        self._app = None
```
schmir: should probably be self.app = None. it least this is what is being used later.

relay/pushservice/pushservice.py
```diff
@@ -0,0 +1,66 @@
...
+        Initializes the push service
+        Args:
+            path_to_keyfile: Path to json keyfile with firebase credentials
+        """
+        cred = credentials.Certificate(path_to_keyfile)
+        self.app = firebase_admin.initialize_app(cred)
+
+    def send_event(self, client_token, event: Event):
+        message = self._build_event_message(client_token, event)
+        messaging.send(message, app=self._app)
```
schmir: We have to make sure that this is compatible with gevent. We don't use any monkeypatching at the moment AFAICT.

requirements.txt
```diff
@@ -19,4 +19,5 @@ tinyrpc
 gevent-websocket
 marshmallow
 flask-sockets
-mypy
\ No newline at end of file
+mypy
+firebase-admin
```
schmir: This should also be added to setup.py's install_requires.

relay/pushservice/pushservice.py
berndbohmeier: Yep that with the app is a mistake.
I can move it into init, but I usually try to have only simple code in init and this seemed to me to much stuff for the init method as I dont know if this need network traffic.

schmir: I see your point. But doing it this way, is putting the burden on the caller, who must now call an additional method in order to get a fully initialized object. And he does, right after creating the object, so the caller doesn't have a need to have 'delayed object initialization'. He could provide all the needed arguments for a full initialization.

If you really need that kind of delayed initialization, I think the class itself should encapsulate it, i.e. it could store the needed parameters for initialization of self._app and create that on the 'first real use'.

But at the moment I think it would be enough to just initialize everything in `__init__`.  It's similar to what is being done now anyway.

relay/pushservice/pushservice.py
schmir: If the library doesn't use a custom C extension, we should be pretty safe with monkey patching.

schmir:
All commits in this PR are also part of PR #114 (just as a note to myself)

relay/pushservice/pushservice.py
berndbohmeier: So monkey patch all at the beginning of the relay.py file?

relay/pushservice/pushservice.py
berndbohmeier: Okay, will do that

PR #115, opened 2018-06-21
Add REST endpoints to add and delete client tokens

### LINKED ISSUE #111: New endpoint to get registered to get push notifications

berndbohmeier:
Add REST endpoints to add and delete client tokens  …
Closes #111

builds on top of #114, so merge that first


berndbohmeier:
I think 404 for a non existing resource should be ok.
Yes, I will add some logic to prevent adding the same token multiple times.

PR #119, opened 2018-06-28
install and use some useful flake8 plugins

schmir:
We install the following flake8 plugins:

- flake8-pep3101: warn about % formatting
- flake8-string-format: warn about errors in .format calls
- flake8-invalid-escape-sequences: warn about invalid escape sequences in
  strings

and flake8-per-file-ignores in order to ignore warnings about % formatting in
one of our source files. flake8 automatically uses those plugins.

These should be pretty uncontroversial.

There are more useful plugins, but they either need work from our side to fix
things or need some discussion.

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -3,6 +3,8 @@ anyconfig==0.9.4
 argh==0.26.2
 attrs==17.4.0
 bitcoin==1.1.42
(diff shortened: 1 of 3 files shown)


berndbohmeier:
LGTM, we can talk about other plugins if you want.

PR #123, opened 2018-06-29
Only send some events as push notification

### LINKED ISSUE #121: Only sent certain Events as Push Notification
Currently we sent all events as push notification. Some should only be send as message. 

Only send 
 - Transfer received,
 - TrustlineUpdate
- TrustlineUpdateRequest received

as notification the rest should be a message.

berndbohmeier:
other events will only be sent as messages because the users
does not need a notification for that
closes #121


berndbohmeier:
For time reasons without review as discussed with ralph

PR #124, opened 2018-06-29
Fix a race condition bug in stream unsubscribe

berndbohmeier:
When inside notify calling client.send a client can unschedule the greenlet and thus delay unsubscription
inside of notify. This means it can be already unsubscribed and we have to handle that


berndbohmeier:
I did not want it to catch all exception and I did not know of another way to check if it is the ValueError for that reason. Will see if I can do it like you suggested.

berndbohmeier:
good catch, this should be a cleaner fix now.

PR #126, opened 2018-07-02
synchronize dependencies in setup.py and requirements.txt

schmir:
Some dependencies had been missing in setup.py. Please not that setup.py depends
on trustlines-contracts, requirements.txt depends on git+https dependency for
that package. Installation of the former won't work, because the package doesn't
live on the public PyPI server.


berndbohmeier:
lgtm

PR #132, opened 2018-07-04
add config file for pre-commit

schmir:
pre-commit [1] is framework for managing git pre-commit hooks. Since I failed to
reformat files with black before committing them in the past, let's try
this. Please note that we don't have any hooks for black enabled at the moment.
But it will run flake8 and do some more useful checks.

After installation of pre-commit, this can be activated by running

  pre-commit install

also fix some things that it already complained about.

[1] https://pre-commit.com/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,20 @@
+repos:
+-   repo: git://github.com/pre-commit/pre-commit-hooks
+    sha: v0.9.1
+    hooks:
+    - id: check-added-large-files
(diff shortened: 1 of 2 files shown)


berndbohmeier:
Should we add pre commit then to the requirements or is that up to the developer to install and enable?

berndbohmeier:
Otherwise LGTM

schmir:
I wouldn't add it to the requirements at the moment. We should probably add a short note to the README if this turns out to be useful.

PR #135, opened 2018-07-04
Exchange events endpoint

### LINKED ISSUE #70: Endpoint for exchange events
Add all exchange related events `LogFill`, `LogCancel`, `LogKill` to new and existing endpoints.

**new endpoints**
`/exchange/:exchangeAddress/events`
`/exchange/:exchangeAddress/users/:userAddress/events`

**existing endpoints**
`/users/:userAddress/events`

dakingha69:
Closes #70


relay/api/app.py
```diff
@@ -74,6 +75,8 @@ def add_resource(resource, url):
     add_resource(ExchangeAddresses, '/exchange/exchanges')
     add_resource(UnwEthAddresses, '/exchange/eth')
     add_resource(OrderDetail, '/exchange/order/<string:order_hash>')
+    add_resource(UserEventsExchange, '/exchange/<address:exchange_address>/users/<address:user_address>/events')
```
berndbohmeier: The other endpoints, for example `/exchange/orderbook` do not separate the different exchanges. Maybe we should then also not do it for the events? Not sure about this, but usually the idea is that it should not matter which exchange was used.

relay/exchange/exchange_db.py
```diff
@@ -137,11 +137,12 @@ def order_filled(self,
         # NOTE old version of web3.py returns bytes as string from contract, so we have to use force_bytes
         order_hash_bytes = force_bytes(order_hash)
         order_orm = self.session.query(OrderORM).filter_by(msg_hash=order_hash_bytes.hex()).first()
-        order_orm.filled_maker_token_amount += filled_maker_token_amount
-        order_orm.filled_taker_token_amount += filled_taker_token_amount
+        if order_orm is not None:
```
berndbohmeier: Did you experience it being None? In what cases can this happen? Maybe we should emit a warning?

relay/exchange/exchange_db.py
```diff
@@ -137,11 +137,12 @@ def order_filled(self,
...
+            order_orm.filled_maker_token_amount += filled_maker_token_amount
+            order_orm.filled_taker_token_amount += filled_taker_token_amount
 
-        if order_orm.to_order().is_filled():
-            self.session.delete(order_orm)
+            if order_orm.to_order().is_filled():
+                self.session.delete(order_orm)
         self.session.commit()
```
berndbohmeier: we only have to commit if we changed something

relay/exchange/exchange_db.py
```diff
@@ -151,9 +152,10 @@ def order_cancelled(self,
         # NOTE old version of web3.py returns bytes as string from contract, so we have to use force_bytes
         order_hash_bytes = force_bytes(order_hash)
         order_orm = self.session.query(OrderORM).filter_by(msg_hash=order_hash_bytes.hex()).first()
-        order_orm.cancelled_maker_token_amount += cancelled_maker_token_amount
-        order_orm.cancelled_taker_token_amount += cancelled_taker_token_amount
+        if order_orm is not None:
```
berndbohmeier: see above

relay/blockchain/exchange_proxy.py
```diff
@@ -49,14 +62,56 @@ def log(log_entry):
...
 
     def start_listen_on_cancel(self, f) -> None:
         def log(log_entry):
             f(log_entry['args']['orderHash'],
               log_entry['args']['cancelledMakerTokenAmount'],
               log_entry['args']['cancelledTakerTokenAmount'])
-        self.start_listen_on(CancelEvent, log)
+        self.start_listen_on(LogCancelEventType, log)
+
+    def get_exchange_events(self,
```
berndbohmeier: I noticed that we have a lot of duplication in those functions. But I will address that in a separate refactoring pull, so in this PR it can stay like that.

relay/api/app.py
dakingha69: Hmm yes I get your point...  My take was to stay with our other event endpoints though:
- `/networks/<address:network_address>/users/<address:user_address>/events`
- `/tokens/<address:token_address>/users/<address:user_address>/events`

relay/exchange/exchange_db.py
dakingha69: The case was when I restarted the `relay` docker container and the order db was empty. But I guess that is not a "normal" case?

relay/exchange/exchange_db.py
dakingha69: ok will change

relay/exchange/exchange_db.py
dakingha69: will change

relay/blockchain/exchange_events.py
```diff
@@ -0,0 +1,55 @@
...
+LogFillEventType = 'LogFill'
+LogCancelEventType = 'LogCancel'
+
+
+class ExchangeEvent(TLNetworkEvent):
+
+    def __init__(self, web3_event, current_blocknumber, timestamp, user=None):
+        super().__init__(web3_event, current_blocknumber, timestamp, from_to_types, user)
+        self.exchange_address = web3_event.get('address')
+        if (is_hex(web3_event.get('args').get('orderHash'))):
```
berndbohmeier: Short comment why this can be hex. Event can be from our database where it is saved as a hex string. If there is an issue, also put a ref to it.

relay/api/app.py
berndbohmeier: Agreed. So we can probably just have both.

PR #148, opened 2018-07-20
Documentation/how to run relay server

schmir:


docs/RelayServer.md
```diff
@@ -0,0 +1,240 @@
...
+     libssl-dev automake autoconf libtool libgraphviz-dev git libpq-dev
+```
+
+The installation instructions assume you create a dedicated user account and put
+files directly into the users home directory. The paths should be adapted to
+your needs.
+
+### parity/geth
+
+You need to run a parity or geth node, which provides the JSONRPC API to the
```
berndbohmeier: Not sure if we should write it that general, afaik we did not test geth. During the PoC if Trustlnes, it did not work using geth because of small differences in the jsonrpc handling. 
We could write that we only officially parity atm, also geth should work and open an issue to verify that geth is working.

docs/RelayServer.md
```diff
@@ -0,0 +1,240 @@
...
+parity. The whole trustlines system may appear to work, but will give wrong
+results if you run with warp mode enabled.
+
+### contracts
+The [trustlines-contracts
+repository](https://github.com/trustlines-network/contracts) contains the
+solidity contracts to be deployed on the blockchain. The ('how to deploy the
+contracts guide')[] <!-- XXX --> contains more information on how to deploy the
+contracts. We assume from now on that the contracts have already been deployed
+and that the `addresses.json` file has been copied to the users home directory.
```
berndbohmeier: This file is not generated when using the command line tool for normal deploying. Usually you will not know all currency networks at the same time. Right now you have to write it manually.

docs/RelayServer.md
```diff
@@ -0,0 +1,240 @@
...
+
+Again we need to clone the git repository, create a virtualenv and install into
+that. Please note that we can't reuse the py-eth-index virtualenv, since both
+projects do have conflicting dependencies.
+```
+cd ~
+git clone https://github.com/trustlines-network/relay
+python3 -m venv ~/opt/relay; ~/opt/relay/bin/pip install -U pip
+cd ~/relay;
+~/opt/relay/bin/pip install -c constraints.txt populus
```
berndbohmeier: Especially this I would say looks for someone not that familiar with venvs and pip as if this is some part of the relay code.

docs/RelayServer.md
```diff
@@ -0,0 +1,240 @@
...
+git clone https://github.com/trustlines-network/py-eth-index
+```
+We need a virtualenv for this repository, so let's create it:
+```
+python3 -m venv ~/opt/py-eth-index; ~/opt/py-eth-index/bin/pip install -U pip
+```
+and install py-eth-index
+```
+cd ~/py-eth-index
+~/opt/py-eth-index/bin/pip install -c constraints.txt -r requirements.txt
```
berndbohmeier: Why do you always specify the full path and not use the venv by activating it? I have never seen that. I find it confusing, also because the location is confusing to me. I have never seen someone putting the venv not in the home folder.

docs/RelayServer.md
schmir: Thanks for the explanation, this wasn't clear to me. I will change it.

docs/RelayServer.md
schmir: Can you please write a short description then? I haven't read the documentation for the contract deployment CLI tools yet.

docs/RelayServer.md
schmir: Sorry, I'm not sure what you mean here. Can you please clarify?

docs/RelayServer.md
schmir: Well, I use ~/opt as some kind of user-writable /opt filesystem here (see https://www.tldp.org/LDP/Linux-Filesystem-Hierarchy/html/opt.html for more information on /opt). 
~/opt lives in the home folder. I am not aware of any location where users should put their virtualenvs. I guess there's no standard location to put them. spotify's dh-virtualenv puts them into /opt.

Activating the virtualenv gives more room to make errors, e.g. you could run one of the pip install commands in the wrong virtualenv (essentially destroying it, since we do have conflicting requirements). 
Though we could create symlinks in ~/bin/ to the ethindex and tl-relay programs, that won't hurt.

docs/RelayServer.md
berndbohmeier: The deploy tool will return the address of the deployed contract. You should describe here how the address.json file is structured and can be written manually if the addresses of the currency network, exchanges, and unwrapping ether contract are known. 
Do you think we need a tool to generate the file?

docs/RelayServer.md
berndbohmeier: ah sorry, I actually did not see that you use ~/opt, I always read /opt.
In the home folder is fine. 
I just think it is distracting to always write ~/opt/env/bin/ethindex and then hard to see what is actually happening.
Just writting ethindex seems to me more clear that we now call the program.

docs/RelayServer.md
berndbohmeier: I think it is hard to see here that you call pip of your virtual env and not something of the relay server, as the executable lives in ~/opt/relay. I think we should name it better so that someone can see that it is a virtual env. So relay env or something like that.

docs/RelayServer.md
schmir: I don't think we need a tool at the moment. Can you please write that description? I think it's easier for you and it doesn't make sense to just describe the json format. We should document how that relates to the contract deployment CLI tools.

berndbohmeier: Ok, will do that tomorrow.

docs/RelayServer.md
schmir: Can we just go on with absolute paths for now? (It's not like it's working). The next time I run through those instructions and will actually run them on a VM, I will change it to use activated virtualenvs.

berndbohmeier: Just how I see it, if you feel it is better this way, we can leave it.

seichris:
looks good

schmir:
I've removed the remark about the unlocked account. AFAICT there are no issues left. So, unless someone  objects, I'm going to merge this later.

Thanks everyone!

PR #151, opened 2018-07-23
add docker-compose based setup

schmir:
This adds a docker-compose based setup and is part of
https://github.com/trustlines-network/relay/issues/140

### DIFF
--- a/docker/trustlines/.env
+++ b/docker/trustlines/.env
@@ -1,0 +1,6 @@
+PGHOST=db
+PGUSER=trustlines
+POSTGRES_USER=trustlines
+PGDATABASE=trustlines
+PGPASSWORD=xxx
+POSTGRES_PASSWORD=xxx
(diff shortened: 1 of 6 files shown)


schmir:
Well, it's just an example, strictly speaking there is no need to keep it updated.

I'm also not quite happy with it and thought about putting the whole docker-compose based setup into a separate repository so we can keep this one clean.

The problem is, there's no place to download it at the moment and if we change the contracts, the one we install while setting up the relay service, will not match the addresses.json, which I also ship.

berndbohmeier:
So the example files and the Docker-compose file are basically there to get fast to a setup that connects you to the current version on l
Kovan right?
Then I agree, it would probably make sense to have that in an own repo.

berndbohmeier:
If we put it in here I think it should include the relay code from the checked out branch and should not contain the contract.json file.

schmir:
I've removed the contracts.json file and added one installation step, which copies the file from the relay image to the README.  `addresses.json` still lives here.
The relay image is now also built from the checked out sources.

So, we could keep that here or move it to a dedicated repo.

PR #153, opened 2018-07-24
Connect DEX and client token db to postgres database.

### LINKED ISSUE #133: Use postgres db for client tokens and exchange
At the moment the client token db only uses a simply file based database. Let it use our postgres database. 
Same for the exchange.

berndbohmeier:
Use BigInteger for now. This might not be enough, but Ill open another issue. 
Closes #133

### DIFF
--- a/relay/exchange/exchange_db.py
+++ b/relay/exchange/exchange_db.py
@@ -1,7 +1,7 @@
 from typing import Tuple, Sequence, Optional
 
 from sqlalchemy.ext.declarative import declarative_base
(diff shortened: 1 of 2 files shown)


berndbohmeier:
I dont think so. I gave up on having it read by them lib and now read the env variables myself. 
Is it important? Do we use them right now?

schmir:
The documentation refers to it and also contains instructions to create a ~/.pgpass file. We may have to adapt the documentation then.

PR #160, opened 2018-07-26
Add internal unsubscription if we know the websocket has disconnected

### LINKED ISSUE #86: Unsubscribe event listener when websocket client disconnects.
We should internally unsubscribe from the event subject when the websocket disconnects. 
Disconnect event calls https://github.com/trustlines-network/relay/blob/develop/relay/api/streams/transport.py#L45

berndbohmeier:
The client keeps a list of subscriptions now so that we can stop listening if the connection is closed
closes #86

### DIFF
--- a/relay/streams.py
+++ b/relay/streams.py
@@ -16,6 +16,7 @@ class Client(object):
 
     def __init__(self) -> None:
         self.subscriptions = []  # type: List[Subscription]
+        self.closed = False
 
     def register(self, subscription: 'Subscription') -> None:
         """
@@ -31,11 +32,23 @@ class Client(object):
         """
         self.subscriptions.remove(subscription)
 
-    def send(self, subscription_id: str, event: Publishable) -> None:
+    def send(self, subscription: 'Subscription', event: Publishable) -> None:
         """
         Sends an event to the client that belongs to a subscription with the given subscription id
         Raises:
             DisconnectedError: This is raised if the client has already disconnected.
+        """
+        if subscription not in self.subscriptions:
+            raise ValueError('Unknown subscription')
+        if self.closed:
+            raise RuntimeError('Client connection is closed')
+        self._execute_send(subscription, event)
+
+    def _execute_send(self, subscription: 'Subscription', event: Publishable) -> None:
+        """
+        Executes the sending
+        Should be implemented by sub class
+        may raise DisconnectedError
         """
         raise NotImplementedError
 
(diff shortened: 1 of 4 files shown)


relay/streams.py
```diff
@@ -12,21 +12,63 @@
...
+    def register(self, subscription: 'Subscription') -> None:
+        """
+        Registers a subscription that this client has done.
+        On closing the connection with `close` these subscription will get unsubscribed
+        """
+        self.subscriptions.append(subscription)
+
+    def unregister(self, subscription: 'Subscription') -> None:
+        """
+        Unregisters a subscription that this client is not listing for anymore.
```
schmir: listing -> listening

relay/streams.py
```diff
@@ -12,21 +12,63 @@
...
+            DisconnectedError: This is raised if the client has already disconnected.
+        """
         raise NotImplementedError
 
+    def close(self) -> None:
+        """
+        Should be called when the connection to this client is closed.
+        This will unsubscribe all registered subscriptions
+        """
+        for subscription in self.subscriptions:
```
schmir: This looks like we may have to iterate over a copy of self.subscriptions. Could that be the case?

berndbohmeier: I dont think so, as we dont delete from that list while iterating. I can still do it to be sure. 
But I think i should probably delete the subscriptions at the end, which I dont do.

schmir: Subscription.unsubscribe calls client.unregister, unregister seems to call self.subscriptions.remove(subscription), so this looks like it is modifying that list. Could also be the case that it's another list. I'm not sure either, what is being modified here.

berndbohmeier: Those are two different lists. A subjects keeps track of its subscription, and a client keeps track where he subscribed to

berndbohmeier: I will copy it to be sure.

berndbohmeier: No wait, actually you are right. My bad.

schmir:
two of the three new test cases seem unrelated, at least they don't fail if the list is not being copied.

Please split such commits next time and add a bit more information.  The reviewer will thank you.

If we fix bugs and write tests for them, I would really like us to write a failing test,commit that, and commit the fix afterwards. Not sure if that works with our CI setup.

I feel like streams.py could/should be simplified. If it was simple enough we wouldn't have the discussions if the list that is being iterrated over, is also being modified or not. But that's probably a different issue.

Nevertheless please push this as is.

PR #168, opened 2018-08-01
use docker images in docker/trustlines

schmir:
please note that we use the latest tag, just like we did previously by using the
develop branch.

see https://github.com/trustlines-network/relay/issues/166

### DIFF
--- a/docker/trustlines/docker-compose.yml
+++ b/docker/trustlines/docker-compose.yml
@@ -32,12 +32,10 @@ services:
     restart: unless-stopped
 
   relay:
+    image: trustlines/relay
     depends_on:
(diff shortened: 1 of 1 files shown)


docker/trustlines/docker-compose.yml
```diff
@@ -32,12 +32,10 @@ services:
     restart: unless-stopped
 
   relay:
+    image: trustlines/relay
```
berndbohmeier: no tag needed here? does it default to latest?

schmir: yes and yes. see commit message for my reasoning.

PR #177, opened 2018-08-03
travis: use build stages

schmir:
We use build stages, so that we have a chance to hide our DOCKER_PASSWORD.

The Dockerfile itself now also uses a staged build and copies the virtualenv to
/opt/relay. The base image itself hasn't changed, i.e. I haven't upgraded to
ubuntu 18.04.

The build is pretty much consistent now with what we're doing in the watch and
py-eth-index repositories.

This will also trigger the deploy stage for tags.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,24 @@
+build
+dist
+*.egg-info
+*.egg/
+*.eggs/
+*.pyc
+*.pyo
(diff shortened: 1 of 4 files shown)


schmir:
This still uses python 3.6 on travis (but that is also the case in the current head of the develop branch)

PR #182, opened 2018-08-08
Add field user and counterParty for consistency with events

berndbohmeier:
Should be backwards compatible as I did not remove the old field.

### DIFF
--- a/relay/api/resources.py
+++ b/relay/api/resources.py
@@ -118,6 +118,8 @@ class Trustline(Resource):
         graph = self.trustlines.currency_network_graphs[network_address]
         data = TrustlineSchema().dump(graph.get_account_sum(a_address, b_address)).data
         data.update({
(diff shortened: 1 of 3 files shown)


relay/api/resources.py
```diff
@@ -118,6 +118,8 @@ def get(self, network_address, a_address, b_address):
         graph = self.trustlines.currency_network_graphs[network_address]
         data = TrustlineSchema().dump(graph.get_account_sum(a_address, b_address)).data
         data.update({
+            'user': a_address,
+            'counterParty': b_address,
             'address': b_address,
```
schmir: address is still being set here. not sure if that's on purpose or if it could be removed.

berndbohmeier: can be removed in the future, I only left it so that this is not a breaking change.

PR #186, opened 2018-08-15
Pypi

schmir:
This implements automatic uploads to pypi via travis. pypi won't allow the use of local version identifiers, so we can only upload tagged releases.


berndbohmeier:
Which part is not allowed, is it only the hash part at the end? Would for example 0.1.1.dev12 work?
We could do that as on develop this is an increasing number.

schmir:
local version identifiers are not allowed, i.e. everything after the `+`:
https://www.python.org/dev/peps/pep-0440/#local-version-identifiers

0.1.1.dev12 should work. but I'd say let's discuss this after your vacation.

PR #190, opened 2018-08-27
Feature/upgrade web3

schmir:


pytest
```diff
@@ -0,0 +1,19 @@
+#! /usr/bin/env python
+
```
berndbohmeier: Would it be possible to rather solve that through conftest.py or is that too late for monkey patching. Needing to run this file instead of pytest is not so nice I think.

relay/exchange/exchange_db.py
```diff
@@ -5,7 +5,7 @@
 from sqlalchemy.orm import sessionmaker
 
 from .order import Order
-from eth_utils import force_bytes
+from relay.compat import force_bytes
```
berndbohmeier: Do we still need this? This was because of a bug in the old web3, so it should not be necessary anymore.

tests/chain_integration/test_exchange_proxy.py
```diff
@@ -209,6 +217,6 @@ def log(order_hash, maker_token_amount, taker_token_amount):
     gevent.sleep(1)
 
     log1 = logs[0]
-    assert log1[0].encode('Latin-1') == order.hash()  # encoding because of bug in web3
+    assert log1[0] == order.hash()  # encoding because of bug in web3
```
berndbohmeier: Please also remove the comment

tests/chain_integration/test_exchange_proxy.py
```diff
@@ -1,15 +1,23 @@
...
 from eth_utils import to_checksum_address
-from ethereum import tester
 
 from relay.blockchain.exchange_proxy import ExchangeProxy
 from relay.constants import NULL_ADDRESS
 from relay.exchange.order import SignableOrder
 
 
+@pytest.fixture
+def accounts(tester, web3):
```
berndbohmeier: We can also remove this here and just alwazs make the make explicit. This is probably better. 
For example
```
maker = tester.a0
taker, *rest = accounts
```

berndbohmeier: or even better probably:
```
maker = order.maker_address
```

pytest
schmir: It might be possible or it might already be too late. I'd rather err on the safe side and put this wrapper around pytest here.

relay/exchange/exchange_db.py
schmir: right. that's one of the things I still wanted to look into. The way I understood the changelog of web3 these calls shouldn't be necessary anymore.

tests/chain_integration/test_exchange_proxy.py
schmir: the accounts fixture here is also something I still wanted to look into. I would have probably introduced a makertaker() fixture that would have returned a tuple.

tests/chain_integration/test_exchange_proxy.py
schmir: sure

tests/chain_integration/test_exchange_proxy.py
schmir: this has been addressed in 8faf063fbf40e703f5e29ba43e729d51b6022d71

PR #192, opened 2018-08-30
Query the nonce from pending transactions

berndbohmeier:
This enables a more up to date nonce information. It will use the pending block to determine the next nonce or on parity it will use the nextNonce function that also includes the pending transactions. At the moment it will always default to parity, as we only use parity anyways. This also requires that the parity api is enabled (--jsonrpc-apis=all or --jsonrpc-apis=parity)

I checked that the e2e tests are still working. To fully test if that now allows to send faster transfers, we need to test it on the staging setup with the app. 

Closes https://github.com/trustlines-network/project/issues/176

### DIFF
--- a/relay/blockchain/node.py
+++ b/relay/blockchain/node.py
@@ -1,13 +1,22 @@
+import logging
 from collections import namedtuple
+
+from relay.logger import get_logger
 
 
 TxInfos = namedtuple('TxInfos', 'balance, nonce, gas_price')
 
 
+logger = get_logger('node', logging.DEBUG)
+
+
 class Node:
 
-    def __init__(self, web3):
+    def __init__(self, web3, is_parity=True):
         self._web3 = web3
+        self.is_parity = is_parity
+        if is_parity:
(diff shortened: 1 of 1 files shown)


relay/blockchain/node.py
```diff
@@ -6,18 +6,24 @@
...
     def transaction_receipt(self, txn_hash):
         return self._web3.eth.getTransactionReceipt(txn_hash)
 
-    def get_tx_infos(self, user_address):
-        return TxInfos(balance=self._web3.eth.getBalance(user_address),
-                       nonce=self._web3.eth.getTransactionCount(user_address),
+    def get_tx_infos(self, user_address, block_identifier='pending'):
+        if self.is_parity and block_identifier == 'pending':
+            nonce = int(self._web3.manager.request_blocking('parity_nextNonce', [user_address]), 16)
+            print('paritynonce')
```
schmir: this should be removed or turned into a logging call

berndbohmeier: Woops my bad. Of course.

PR #198, opened 2018-09-13
Increase requestEth threshold

jensneuber:
Increase the request eth threshold to 5


jensneuber:
To be able to test the app. There is no way to get Eth if balance is 0.000001

PR #199, opened 2018-09-14
Read the feeDivisor value from contract

cducrest:
Read the feeDivisor value from contract when generating network instead of using 100.
Related to https://github.com/trustlines-network/relay/issues/194

### DIFF
--- a/relay/blockchain/currency_network_proxy.py
+++ b/relay/blockchain/currency_network_proxy.py
@@ -10,6 +10,8 @@ import gevent
 import relay.concurrency_utils as concurrency_utils
 from .proxy import Proxy, reconnect_interval, sorted_events
 from relay.logger import get_logger
+from web3.exceptions import BadFunctionCallOutput, ValidationError
+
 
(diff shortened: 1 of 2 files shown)


relay/blockchain/currency_network_proxy.py
```diff
@@ -52,6 +54,10 @@ def __init__(self, web3, abi, address: str) -> None:
         self.name = self._proxy.call().name().strip('\0')  # type: str
         self.decimals = self._proxy.call().decimals()  # typ: str
         self.symbol = self._proxy.call().symbol().strip('\0')  # type: str
+        try:
+            self.capacityImbalanceFeeDivisor = self._proxy.call().capacityImbalanceFeeDivisor()  # type: str
```
schmir: why use `# type: str` hint? that's misleading, should be an integer.

cducrest:
I resolved the problem shown by travis by reverse-engineering, and now the travis checks pass but I cannot run the relay server anymore. I am not sure who's correct.

PR #201, opened 2018-09-17
Implements the interests on the relay server

cducrest:
https://github.com/trustlines-network/project/issues/202
- The endpoints were modified to include information about interests.
- The balance of accounts are directly modified whenever queried to take into account interests.
- The different pathfinding take into account interests
- Not sure if I should use `data['interest_ab']` or `data['interest_ba']` in each of my calculation, since this abstraction makes no sense to me. It appears that `creditline_ba` is the credit limit given by b to a (a can spend up to creditline_ba) so I will assume `interest_ba ` is the interest given by b to a.

- Did not implement safeRippling yet, as I want to be sure of the contract version first.
- If we modify the calculus for interest on smart contract, we will have to modify it in "interests.py" too.


relay/network_graph/interests.py
```diff
@@ -0,0 +1,25 @@
...
+    # apparently not, need to refactor
+    new_time = time.time()
+
+    # I assume interest_ba is the interest given by b to a. This might be wrong.
+    if a < b:
+        if data['balance_ab'] > 0:  # b owes a
+            interest = calculate_interests(data['balance_ab'], data['interest_ab'], data['m_time'], new_time)
+        else:  # a owes b
+            interest = calculate_interests(data['balance_ab'], data['interest_ba'], data['m_time'], new_time)
+    else:
```
schmir: the if and then else block of the outer if statement are the same, which make me think there is an error.

relay/network_graph/interests.py
```diff
@@ -0,0 +1,25 @@
+import time
+
+def calculate_interests(balance, rate, old_time, new_time):
+    return balance * (new_time - old_time) / (60 * 60 * 24 * 365) * rate / 100000
+
+def apply_interests(a, b, data):
+    """Applies the interest to the balance in data. The interpretation of data depends on whether a < b.
+    a and b play symmetrical roles"""
+    # apparently not, need to refactor
+    new_time = time.time()
```
schmir: I'd like to see the time get passed in as an argument. Also isn't data['m_time'] in days and time.time() is in seconds?

schmir:
One thing that is disturbing and surprising to me is the fact that the graph is modified while calling a function like find_path. I wouldn't expect such a function to change the graph. The same is true for the balance property, but with the balance property it's at least clear which part of the graph is modified.

To me there's no way to reason about which parts of the graph are modified when calling find_path (which may not even be able to find a path, yet it would change the graph).

AFAICT the contracts use days in mtime, so we could apply interests just once a day for the whole graph.

Even if that wouldn't work, one could compute a list of interests/updated balances for each account while traversing the graph without modifying it. That would also be a cleaner solution.

relay/network_graph/interests.py
cducrest: Everything should be in seconds, the smart contract uses seconds to calculate interests (at least in the interest branch).

relay/network_graph/interests.py
cducrest: I do not think there is an error, the meaning of the data depends on the point of view of the data (a<b or b<a). But only one data is stored for both point of views. I believe this is confusing, which is why I kept both if and else statements even though the code is the same (easier to reason about for me).

cducrest:
Done with required modifications
- Still no safe rippling strategy
- Still need to modify the interests calculations when modified in the smart contract

docs/RelayAPI.md
```diff
@@ -99,14 +99,20 @@ curl https://relay0.testnet.trustlines.network/api/v1/networks/0xC0B33D88C704455
...
   "abbreviation": "HOU",
   "address": "0xC0B33D88C704455075a0724AA167a286da778DDE",
   "decimals": 2,
   "name": "Hours",
-  "numUsers": 3
+  "numUsers": 3,
+  "defaultInterests": 1000,
+  "customInterests": false,
+  "safeInterestRippling": false
 }
```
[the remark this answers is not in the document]
cducrest: We could, but the interestRateDecimals is fixed so that interests are expressed in 0.01% per year. It cannot be modified, so does it make sense to have it in the endpoint?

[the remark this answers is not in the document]
dakingha69: Ahh ok... but are there plans in the future to make it configurable? I think @berndbohmeier mentioned it at some point. But I would just include them for now so we could change it later. Will also ping Bernd to get some input from him about this.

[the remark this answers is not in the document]
berndbohmeier: I can include that, so that we can change it later.

schmir:
`resources.NetworkList.get` should use the `dump_result_with_schema decorator`. We still have other locations in that file where we could use the decorator, but calling marshmallows dump for each element in a list is a worse offender.

And I've got a question: Why don't we use continuous compounding, when computing interests? 

Other than that it looks fine to me.

schmir:
The Account.balance setter doesn't take the `mtime` into account, while the getter computes interests.
I think this is problematic, since for an Account object a,  running

```
a.balance = a.balance
```
shouldn't change the balance. AFAICT, this might increase/decrease the balance with every assignment.

PR #205, opened 2018-09-27
Refactor/web3update

berndbohmeier:
Remove the ETHINDEX flag as it is now the default
Use hexbytes datatype internally instead of removing it
Remove tester mock class and instead use a testaccount
Refactored call() and transact to new web3 v4 pattern

### DIFF
--- a/relay/exchange/order.py
+++ b/relay/exchange/order.py
@@ -139,8 +139,8 @@ class SignableOrder(Order):
                          expiration_timestamp_in_sec,
                          salt,
                          v=0,
-                         r=hexbytes.HexBytes('0x00'),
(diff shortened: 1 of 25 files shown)


relay/exchange/order.py
```diff
@@ -138,11 +139,11 @@ def __init__(
                          expiration_timestamp_in_sec,
                          salt,
                          v=0,
-                         r=b'',
-                         s=b'')
+                         r=hexbytes.HexBytes('0x00'),
+                         s=hexbytes.HexBytes('0x00'))
```
schmir: r and b have been empty before, now they contain the zero byte AFAICT. Was that intended?

PR #208, opened 2018-10-22
Feature/possible triangulations

schmir:
This adds the CloseTrustline resource (which still needs to be documented), enhancements to the way we use marshmallow and enhancements for running the tests.

### DIFF
--- a/relay/api/resources.py
+++ b/relay/api/resources.py
@@ -12,18 +12,14 @@ from marshmallow import validate
 from relay.utils import sha3
 from relay.blockchain.currency_network_proxy import CurrencyNetworkProxy
 from relay.blockchain.unw_eth_proxy import UnwEthProxy
-from relay.blockchain.unw_eth_events import UnwEthEvent
-from relay.blockchain.exchange_events import ExchangeEvent
-from relay.blockchain.currency_network_events import CurrencyNetworkEvent
 from relay.api import fields as custom_fields
 from .schemas import (CurrencyNetworkEventSchema,
                       UserCurrencyNetworkEventSchema,
-                      UserTokenEventSchema,
-                      ExchangeEventSchema,
                       AccountSummarySchema,
                       ExtendedAccountSummarySchema,
                       TxInfosSchema,
-                      PaymentPathSchema)
+                      PaymentPathSchema,
+                      AnyEventSchema)
 from relay.relay import TrustlinesRelay
 from relay.concurrency_utils import TimeoutException
 from relay.logger import get_logger
@@ -232,14 +228,15 @@ class UserEvents(Resource):
     }
 
     @use_args(args)
+    @dump_result_with_schema(AnyEventSchema(many=True))
     def get(self, args, user_address: str):
         type = args['type']
         from_block = args['fromBlock']
         try:
-            events = self.trustlines.get_user_events(user_address,
-                                                     type=type,
-                                                     from_block=from_block,
-                                                     timeout=self.trustlines.event_query_timeout)
+            return self.trustlines.get_user_events(user_address,
+                                                   type=type,
+                                                   from_block=from_block,
+                                                   timeout=self.trustlines.event_query_timeout)
         except TimeoutException:
             logger.warning(
                 "User events: event_name=%s user_address=%s from_block=%s. could not get events in time",
@@ -247,15 +244,6 @@ class UserEvents(Resource):
                 user_address,
                 from_block)
             abort(504, TIMEOUT_MESSAGE)
-        serialized_events = []
-        for event in events:
-            if isinstance(event, CurrencyNetworkEvent):
-                serialized_events.append(UserCurrencyNetworkEventSchema().dump(event).data)
-            if isinstance(event, UnwEthEvent):
-                serialized_events.append(UserTokenEventSchema().dump(event).data)
-            if isinstance(event, ExchangeEvent):
-                serialized_events.append(ExchangeEventSchema().dump(event).data)
-        return serialized_events
 
 
 class EventsNetwork(Resource):
--- a/relay/api/resources.py
+++ b/relay/api/resources.py
@@ -329,6 +329,18 @@ class RequestEther(Resource):
         return self.trustlines.node.send_ether(address)
 
 
+def _estimate_gas_for_transfer(trustlines: TrustlinesRelay,
+                               payment_path: PaymentPath,
+                               network_address: str) -> PaymentPath:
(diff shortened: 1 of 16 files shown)


relay/api/resources.py
```diff
@@ -416,24 +409,51 @@ def post(self, args, network_address: str):
...
+
+        graph = self.trustlines.currency_network_graphs[network_address]
+        balance = graph.get_balance(source, target_reduce)
+        value = -balance
+
+        if balance == 0:
+            return PaymentPath(fee=0, path=[], value=value, estimated_gas=0)
+
+        if balance > 0:
+            raise RuntimeError("balance is positive. Cannot reduce debt in CloseTrustline resource.")
```
berndbohmeier: return client error code instead.. I think 409 conflict is probably best.

berndbohmeier: either with abort or own exception

relay/api/schemas.py
```diff
@@ -68,6 +73,26 @@ class UserExchangeEventSchema(ExchangeEventSchema):
...
+class AnyEventSchema(OneOfSchema):
+    type_schemas = {
+        'CurrencyNetworkEvent': UserCurrencyNetworkEventSchema,
+        'UnwEthEvent': UserTokenEventSchema,
+        'ExchangeEvent': ExchangeEventSchema,
+    }
+
+    type_field = '__class__'
+
+    def get_obj_type(self, obj):
```
berndbohmeier: I think this is the default implementation, so you dont have to override this method I guess?

relay/blockchain/currency_network_proxy.py
```diff
@@ -198,3 +198,16 @@ def get_all_network_events(self,
...
+            return 0
+        source = payment_path.path[0]
+        target = payment_path.path[-1]
+        try:
+            return self._proxy.estimateGas({'from': source}).transfer(target,
+                                                                      payment_path.value,
+                                                                      payment_path.fee,
+                                                                      payment_path.path[1:])
+        except ValueError as e:  # should mean out of gas, so path was not right.
+            return 0
```
berndbohmeier: I dont think is is a good idea to handle that error with returning 0. This is hiding the error and the value is confusing. It is out of gas so if we would return a value the suitable would be infinity. I think it is better to throw an exception like InvalidPathException or OutOfGasException

relay/api/resources.py
```diff
@@ -333,6 +329,18 @@ def post(self):
...
 
 
+def _estimate_gas_for_transfer(trustlines: TrustlinesRelay,
+                               payment_path: PaymentPath,
+                               network_address: str) -> PaymentPath:
+    proxy = trustlines.currency_network_proxies[network_address]
+    payment_path.estimated_gas = proxy.estimate_gas_for_payment_path(payment_path)
+
+    if payment_path.estimated_gas == 0:
+        return PaymentPath(fee=0, path=[], value=payment_path.value, estimated_gas=0)
```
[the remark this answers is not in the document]
berndbohmeier: Okay found it, but I think this is bad and confusing.

relay/api/schemas.py
schmir: I'm using the exact logic that has been in use before. It differs from the default implementation because it's using isinstance checks and will return the same object type for all subclasses.

relay/blockchain/currency_network_proxy.py
schmir: Well, catching the error is exactly why I've introduced this function. Returning 0 is also consistent with what we do in our REST API. I can understand your argument, especially given the fact that this has been moved to another file. I can move it back to resources.py as a helper function or add a comment explaining it.
Raising another error doesn't help the code in resources calling this method.

relay/api/schemas.py
berndbohmeier: what I mean was that AFAIK this is the same what the default logic of get_object_type is doing so I would just remove the function.

relay/api/resources.py
schmir: I think 501 not implemented is more appropriate. Where's the conflict?

relay/api/schemas.py
schmir: what I tried to eplain is that this is not the case. It differs from the default implementation.

relay/blockchain/currency_network_proxy.py
berndbohmeier: No it is not the same. What we do in the other function only makes sense because we return an empty path which means that no path was found, so no payment can be made. Here we return the value for a certain path, which is totally different. 0 with empty path is consistent, 0 with a path that throws is not.

relay/api/resources.py
berndbohmeier: We can also go with 501. It depends if we think it is our fault that it is not implemented or theirs for using it in the wrong way. 
The conflict would be that the network is not in a state that it should be when calling this endpoint.

relay/blockchain/currency_network_proxy.py
berndbohmeier: I think raising an error would make it clear and in the calling code you could just catch it and do the same logic where there is now the == 0 check.

relay/api/schemas.py
berndbohmeier: Okay maybe I am wrong. It looks like you just return the class name and on the webseite they state this is the default implementation
> By default get_obj_type() returns obj.__class__.__name__, so you can just reuse that to save some typing:

relay/blockchain/currency_network_proxy.py
schmir: > What we do in the other function only makes sense because we return an empty path which means that no path was found, so no payment can be made

This is just how you interpret that empty path, i.e.you give a special meaning to that return value.

We could as well raise an error instead of returning a special value in the REST API.

Anyway,I've moved the exception catching to the calling function in 50dfea897dd67d310c0d74fb55c42.

relay/api/schemas.py
schmir: > Okay maybe I am wrong. It looks like you just return the class name and on the webseite they state this is the default implementation

The result differs for a BalanceUpdateEvent, which is a subclass of CurrencyNetworkEvent.

This doesn't mean that we couldn't or shouldn't use the default implementation, just that it differs.

Maybe dispatching based on the type field would be better. I have to take a second look at it. Ideally we would be able to load that data again with marshmallow. This is currently not possible AFAICT.

schmir: opened in an issue here: https://github.com/trustlines-network/relay/issues/209

PR #210, opened 2018-10-26
Fix/misc graph stuff

schmir:
Here are some enhancements, especially for the `find_path_triangulation` function (mostly about removing code not needed and about improving readability).

@cducrest That's why I like you to review it.


relay/network_graph/dijkstra_weighted.py
```diff
@@ -127,6 +127,15 @@ def get_capacity(u, v, data):  # gets the capacity from u to v
             "node %s not reachable from %s" % (source, target))
 
 
+def get_balance(G, a, b):
```
cducrest: Why do we put this outside of "find_path_triangulation" if we do not use it elsewhere?

PR #216, opened 2018-11-14
Change interest calc to continous compounding

berndbohmeier:
The interests are now calculated with a taylor approximation of
continous compounding to match the calculation in the contracts
Also add some tests to test that the calculation is the same as in the
smart contracts

- During that I refactored some parts of the code:

- Refactor trustline access

It was a hugh mess how we accessed the trustline data. I tried to
clean it up a bit with some getter setter functions

- Refactor call to time.time() to be now within graph.py.
    This makes the code easier to reuse and test
    We should probably refactor that further and move it even more up so
    that it is within the trustlinesrelay so that the cached graph only
    dependes on its own state.


relay/network_graph/graph.py
```diff
@@ -334,28 +312,35 @@ def dump(self):
...
         assert post_balance <= pre_balance
         if -post_balance > creditline:
             return None  # no valid path
         return cost
 
+    def _get_capacity(self, data, u, v):  # gets the capacity from u to v
+        balance_with_interests = self._get_balance(data, u, v)
+        return get_creditline(data, v, u) + balance_with_interests
+
+    def _get_balance(self, data, u, v):
```
cducrest: Should we rename that function to represent that it gets the balance + interest on current server time?
Something like "_get_balance_with_interests_on_current_time()" or "_get_balance_with_interests()"

berndbohmeier: yeah probably makes sense

PR #218, opened 2018-11-15
add test case that demonstrates a bug in dijkstra_weighted.find_path

schmir:
Here's a failing test case that demonstrates a bug in
dijkstra_weighted.find_path.

Do not merge yet.

see issue #219


schmir:
I've started a new impementation of the path finding in the `fix/path-finding` branch. Therefore I'm closing this one. 
The minimal test case is squash merged in ef17de6528d3a92519360296de9b516ce591b45f

PR #221, opened 2018-11-25
Fix/path finding

### LINKED ISSUE #219: `dijkstra_weighted.find_path` returns path with wrong cost
Commit d94e3220e839dc0da5f918a1e049d503d6241cee adds a test case that demonstrates the bug.

Running that file results in the following output:

```
Test session starts (platform: linux, Python 3.6.6, pytest 3.8.0, pytest-sugar 0.9.1)
rootdir: /home/ralf/trustlines/relay, inifile: setup.cfg
plugins: sugar-0.9.1, cov-2.6.0


――――――――――――――――――――――――――――――――――――――――――――――――――――――――――― test_find_path_cost_wrong_bug ―――――――――――――――――――――――――――――――――――――――――――――――――――――――――――

graph = <networkx.classes.graph.Graph object at 0x7f6b66d49710>

    def test_find_path_cost_wrong_bug(graph):
        """our current implementation fails to compute the correct cost for some
        paths"""
        cost_path = find_path(
            graph, source=5, target=32, get_fee=lambda e, u, v, d: e["fee"], value=0
        )
>       sanity_check_fees(graph, cost_path)

test_find_path.py:42: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

graph = <networkx.classes.graph.Graph object at 0x7f6b66d49710>, cost_path = (500, [5, 15, 16, 17, 18, 19, ...])

    def sanity_check_fees(graph, cost_path):
        print(f"Checking {cost_path}")
        cost, path = cost_path
    
        sum_fees = 0
        for source, target in zip(path, path[1:]):
            fee = graph[source][target]["fee"]
            sum_fees += fee
            print(f"{source} -> {target}  fee={fee} sum_fees={sum_fees}")
>       assert sum_fees == cost, f"cost for this path is wrong {cost_path}"
E       AssertionError: cost for this path is wrong (500, [5, 15, 16, 17, 18, 19, 32])
E       assert 105 == 500

test_find_path.py:33: AssertionError
--------------------------------------------------------------- Captured stdout call ----------------------------------------------------------------
Checking (500, [5, 15, 16, 17, 18, 19, 32])
5 -> 15  fee=1 sum_fees=1
15 -> 16  fee=1 sum_fees=2
16 -> 17  fee=1 sum_fees=3
17 -> 18  fee=1 sum_fees=4
18 -> 19  fee=1 sum_fees=5
19 -> 32  fee=100 sum_fees=105

 tests/unit/network_graph/test_find_path.py ⨯                                                                                         100% ██████████

Results (0.13s):
       1 failed
         - tests/unit/network_graph/test_find_path.py:36 test_find_path_cost_wrong_bug
```

schmir:
This adds a new, clean implementatio of the path finding algorithm, where the cost function can be parametrized.
The old find_path is implemented in terms of the new function.

fixes #219 

We should be able to use this function for closing trustlines later, since it allows multi-source/multi-target path finding.

### DIFF
--- a/relay/network_graph/alg.py
+++ b/relay/network_graph/alg.py
@@ -1,0 +1,146 @@
+"""graph algorithms"""
+
+import heapq
+from typing import List, Set, Dict, Callable, Iterable
+import networkx as nx
+import abc
+
+
+class CostAccumulator(metaclass=abc.ABCMeta):
+    @abc.abstractmethod
+    def zero(self):
+        """return 'zero cost' element, which is the initial cost for a one-node path
+
+        This must be less than or equal to any value
+        total_cost_from_start_to_dst returns.
+        """
+        pass
+
+    @abc.abstractmethod
+    def total_cost_from_start_to_dst(
+        self, cost_from_start_to_node, node, dst, edge_data
+    ):
+        """
+        compute the total cost from one of the starting nodes to dst via node
+        given that the cost from one of the starting nodes to node is the given parameter
+        cost_from_start_to_node.
+
+        In other words this adds the cost for the hop from node to dst to
+        cost_from_start_to_node.
+
+        edge_data is the data that's stored inside the graph for the edge from
+        node to dst.
+
+        This function must return a value greater or equal to the given
+        cost_from_start_to_node parameter when used with the least_cost_path
+        function below, since that implement dijkstra's graph finding
+        algorithm, where negative weights are not allowed.
+
+        It may also return None, which means that this path - from one of the
+        starting nodes, to node, to dst - is forbidden.
+        """
+        pass
+
+    def compute_cost_for_path(self, graph: nx.graph.Graph, path: List):
+        cost = self.zero()
+        for source, dst in zip(path, path[1:]):
+            cost = self.total_cost_from_start_to_dst(
+                cost, source, dst, graph.get_edge_data(source, dst)
+            )
+        return cost
+
+
+def _build_path_from_backlinks(dst: List, backlinks: Dict):
+    path = [dst]
+    while True:
+        dst = backlinks[dst]
+        if dst is None:
+            path.reverse()
+            return path
+        path.append(dst)
+
+
+def _least_cost_path_helper(
+    graph: nx.graph.Graph,
+    target_nodes: Set,
+    queue: List,
+    least_costs: Dict,
+    backlinks: Dict,
+    cost_fn: Callable,
+    max_cost=None
+    #    node_filter,
+    #    edge_filter,
+):
+    graph_adj = graph.adj
+
+    while queue:
+        cost_from_start_to_node, node = heapq.heappop(queue)
+        if node in target_nodes:
+            return cost_from_start_to_node, _build_path_from_backlinks(node, backlinks)
+
+        if cost_from_start_to_node > least_costs[node]:
+            continue  # we already found a cheaper path to node
+
+        for dst, edge_data in graph_adj[node].items():
+            cost_from_start_to_dst = cost_fn(
+                cost_from_start_to_node, node, dst, edge_data
+            )
+            if cost_from_start_to_dst is None:  # cost_fn decided this path is forbidden
+                continue
+
+            if max_cost is not None and max_cost < cost_from_start_to_dst:
+                continue
+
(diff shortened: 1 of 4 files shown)


relay/network_graph/alg.py
```diff
@@ -0,0 +1,129 @@
+"""graph algorithms"""
+
+import heapq
+from typing import List, Set, Dict, Callable
+import networkx as nx
+import abc
+
+
+class CostAccumulator(metaclass=abc.ABCMeta):
```
berndbohmeier: Nice!

relay/network_graph/alg.py
```diff
@@ -0,0 +1,129 @@
...
+    def zero(self):
+        """return 'zero cost' element, which is the initial cost for a one-node path
+
+        This must be less than or equal to any value
+        accumulate_cost_for_next_hop returns.
+        """
+        pass
+
+    @abc.abstractmethod
+    def accumulate_cost_for_next_hop(
```
berndbohmeier: Took me some time to understand this, what costs come back, what is graph data, maybe we can rename it and name the arguments better?
for example what about 
`total_cost_from_start_to_dst(self, cost_from_start_to_node, node, dst, edge_data_node_dst)`

relay/network_graph/alg.py
```diff
@@ -0,0 +1,129 @@
...
+                or cost_from_start_to_dst < least_cost_found_so_far_from_start_to_dst
+            ):
+                heapq.heappush(queue, (cost_from_start_to_dst, dst))
+                least_costs[dst] = cost_from_start_to_dst
+                backlinks[dst] = node
+
+    raise nx.NetworkXNoPath("no path found")
+
+
+def least_cost_path(
```
berndbohmeier: I would really like to get a hint, that this is dijkstra, so that a future programmer how wants to change it, knows what he needs to look for. So either rename this function or have another abstraction layer that this function calls `dikstra_multipath` or similar. 
If you dont like either, we should at least hint in a comment, that this is Dijkstra

relay/network_graph/alg.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        compute the total cost from one of the starting nodes to dst via node
+        given that the cost from one of the starting nodes to node is
+        cost_from_start_to_node.
+
+        In other words this adds the cost for the hop from node to dst to
+        cost_from_start_to_node
+
+        This function must return a value greater or equal to the given
+        cost_from_start_to_node parameter.
```
berndbohmeier: We should also hint in the documentation that this method can return null to signal that this is a forbidden path.

relay/network_graph/alg.py
```diff
@@ -0,0 +1,129 @@
...
+
+        if cost_from_start_to_node > least_costs[node]:
+            continue  # we already found a cheaper path to node
+
+        for dst, edge_data in graph_adj[node].items():
+            cost_from_start_to_dst = cost_fn(
+                cost_from_start_to_node, node, dst, edge_data
+            )
+            if cost_from_start_to_dst is None:
+                continue
```
berndbohmeier: comment what this means

tests/unit/network_graph/test_find_path.py
```diff
@@ -0,0 +1,38 @@
...
+
+    sum_fees = 0
+    for source, target in zip(path, path[1:]):
+        fee = graph[source][target]["fee"]
+        sum_fees += fee
+        print(f"{source} -> {target}  fee={fee} sum_fees={sum_fees}")
+    assert sum_fees == cost, f"cost for this path is wrong {cost_path}"
+
+
+def test_find_path_cost_wrong_bug(graph):
```
berndbohmeier: link to issue and update comment to reflect that this is a regression test

relay/network_graph/alg.py
berndbohmeier: hint to forbidden path or similar

PR #224, opened 2018-11-26
Feature/new close trustline

schmir:
This implements the close trustline functionality with the new path finding function.

I  still have to put this through the end2end tests. But I think it's already ready to be reviewed.

The gas cost estimation in the CloseTrustline reesource still has to be adapted AFAICT, but that was also true before.

### DIFF
--- a/relay/blockchain/currency_network_proxy.py
+++ b/relay/blockchain/currency_network_proxy.py
@@ -202,7 +202,7 @@ class CurrencyNetworkProxy(Proxy):
         return sorted_events(list(itertools.chain.from_iterable(results)))
 
     def estimate_gas_for_transfer(self, sender, receiver, value, max_fee, path):
-        return self._proxy.estimateGas({'from': sender}).transfer(receiver, value, max_fee, path)
+        return self._proxy.functions.transfer(receiver, value, max_fee, path).estimateGas({'from': sender})
 
     def estimate_gas_for_payment_path(self, payment_path):
         """estimate gas for doing a transfer for the given payment_path"""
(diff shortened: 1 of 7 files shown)


schmir:
After having been brainwashed by meetings today, I think I found a small problem:

The call to `cost_accumulator.compute_cost_for_path` may fail if `max_fees` doesn't allow the two hops from/to `source` to be added to the path found. The code currently doesn't handle that. I'll provide a fix for that.

relay/blockchain/currency_network_proxy.py
```diff
@@ -214,3 +214,10 @@ def estimate_gas_for_payment_path(self, payment_path):
                                                                   payment_path.value,
                                                                   payment_path.fee,
                                                                   payment_path.path[1:])
+
+    def estimate_gas_for_close_trustline(self, source, other_party, max_fee, path):
+        """estimate gas for doing a transfer for the given payment_path"""
+        return self._proxy.estimateGas({'from': source}).closeTrustlineByTriangularTransfer(
```
berndbohmeier: this is the deprecated way right? We should use _proxy.functions.<method name>.estimateGas

relay/network_graph/graph.py
```diff
@@ -154,6 +154,138 @@ def __init__(self,
...
+            get_interest_rate(edge_data, dst, node),
+            self.timestamp - get_mtime(edge_data))
+
+        fee = imbalance_fee(self.capacity_imbalance_fee_divisor, pre_balance, self.value - sum_fees)
+
+        if sum_fees + fee > self.max_fees:
+            return None
+
+        # check that we don't exceed the creditline
+        capacity = pre_balance + get_creditline(edge_data, node, dst)
```
berndbohmeier: This is the wrong creditline. The creditline used here needs to be the one given by dst to node, so `get_creditline(dst, node)`

berndbohmeier: Maybe add a test to be sure.

PR #234, opened 2018-12-11
upgrade dev tools dependencies

schmir:
[PR against the fix/adapt-to-contracts-0.4 branch]

upgrade flake8, pytest and mypy. Fix new warnings detected by newer flake8.

pre-commit already uses a newer flake8 version and so I'd like to see the same
errors in our setup.

This also removes CurrencyNetworkProxies handling of missing functions and
attributes in the contract. We've had several breaking changes in the contracts
and can't expect it to work with the old contracts anyway.


schmir:
already merged to develop.

PR #235, opened 2018-12-12
pass timestamp to graph methods

schmir:
get_account_sum, find_path and find_maximum_capacity_path take a timestamp
argument.

I couldn't make timestamp a required keyword argument, because mypy wont't let
me. I'm waiting for the dev tools update branch to be merged.

See https://github.com/trustlines-network/relay/issues/222


schmir:
This is already merged to develop. Not sure why github doesn't recognize it.

PR #236, opened 2018-12-28
Feature/better max capacity calculations

### LINKED ISSUE #211: Improve max capacity path calculations
https://github.com/trustlines-network/mobileapp/issues/296

Improve calculations to make the result returned better.

### LINKED ISSUE #227: Find max capacity without fee on last hop
See https://github.com/trustlines-network/project/issues/267

cducrest:
Improve the max capacity path calculations to the point where it should give the exact max sendable amount.
It made no sense regarding tests to improve it without updating the estimation to match with the new fee specification, so this pull request also includes the change from estimation of fee following `floor(imbalance) + 1` and fee on single hop transactions to `ceiling(imbalance)` and no fee on single hop tx.

Closes #211
Closes #227
Related to https://github.com/trustlines-network/mobileapp/issues/296


schmir:
Does this give the exact result now? I have an implementation using a binary search to find the exact value.

cducrest:
I had a binary search at some point too but I figure it gives the exact result without the binary search. Which makes sense now that we changed the fee specification to make it (almost) symmetrical.
Did you push somewhere your implementation and I failed to see it?

schmir:
no,  I didn't push it until now. It's now in the `feature/better_max_capacity_calculations` branch.
The relevant commit is 38f144d1b81f79d11bb87a8e10e9f9e4cdede761.

I'm fine with going with your version. Maybe we could use some of the test functions from my version.

berndbohmeier:
I added some tests to demonstrate that the current implementation is wrong.
I also added a much simpler implementation that will get us the right results I believe. Please verify.

tests/unit/network_graph/test_graph.py
```diff
@@ -284,39 +275,161 @@ def test_capacity_path_multi_hops_positive_balance(complex_community_with_trustl
     assert value == 60000
 
 
+def test_capacity_path_single_hop_with_fees(complex_community_with_trustlines_and_fees):
```
berndbohmeier: The following tests are all the same only with different data. This is not acceptable, this is so much code that does not add anything but makes more work to review, and to maintain. We have to cut that down by reusing the code only feeding it with the different test data.

tests/unit/network_graph/test_graph.py
```diff
@@ -284,39 +275,161 @@ def test_capacity_path_multi_hops_positive_balance(complex_community_with_trustl
...
-    capacity, path = complex_community_with_trustlines_and_fees.find_maximum_capacity_path(
-                A, B)
+    source = A
+    destination = B
 
-    value, path = complex_community_with_trustlines_and_fees.find_path(A, B, capacity)
-    assert path == [A, B]
+    sendable, max_path = complex_community_with_trustlines_and_fees.find_maximum_capacity_path(source, destination)
+    assert max_path == [A, B]
+    print(sendable)
```
berndbohmeier: Remove print statement

schmir:
@berndbohmeier very nice.

I've got some small suggestions and questions.

Suggestions:

 `CapacityAccumulator` could be renamed to `SenderPaysCapacityAccumulator` analogous to `SenderPaysCostAccumulatorSnapshot`.  At least the docstring in `CapacityAccumulator` should mention the fact that it works in this direction. `last_hop_fee` could be renamed `previous_hop_fee`. 

There's some code duplication in the tests for checking and making sure the path is the best path. I could live with that, not sure if you can :)

Question:

Why don't we walk the graph in reverse order like we do when using the `SenderPaysCostAccumulatorSnapshot` ? I could probably think more about that, but that might make my head hurt and you already may have an answer. 

Observation

In `ReceiverPaysCostAccumulator` we use `(sum_fees, previous_hop_fee, num_hops)` and not like in your code `(sum_fees, num_hops, previous_hop_fee)`. We should probably switch to using the latter, since the `previous_hop_fee` will not get paid out for the last hop, but the number_of_hops should matter more.

cducrest:
It works.
The logic behind the algorithm is proper.

relay/network_graph/graph.py
```diff
@@ -296,35 +295,46 @@ def total_cost_from_start_to_dst(
...
 
     def zero(self):
-        return (-math.inf, 0)  # We use (- capacity, num_hops) as cost
+        return (-math.inf, 0, 0)  # We use (- capacity, num_hops, last_hop_fee) as cost
 
     def total_cost_from_start_to_dst(
         self, cost_from_start_to_node, node, dst, edge_data
     ):
 
         capacity_from_start_to_node = - cost_from_start_to_node[0]
```
cducrest: Can we not use a namedtuple for the cost to make it more clear than cost[0], cost[1], cost [2].

relay/network_graph/graph.py
```diff
@@ -296,35 +295,46 @@ def total_cost_from_start_to_dst(
...
             get_interest_rate(edge_data, dst, node),
             self.timestamp - get_mtime(edge_data))
-        return balance + get_creditline(edge_data, dst, node)
+
+    def get_capacity(self, node, dst, edge_data):
+        return self.get_balance(node, dst, edge_data) + get_creditline(edge_data, dst, node)
 
     def zero(self):
-        return (-math.inf, 0)  # We use (- capacity, num_hops) as cost
+        return (-math.inf, 0, 0)  # We use (- capacity, num_hops, last_hop_fee) as cost
```
cducrest: I feel like this is a hack to put the last_hop_fee in the cost as we do not actually try to minimise it, we just want to have access to it when calculating the capacity of the next edge.

tests/unit/network_graph/test_graph.py
berndbohmeier: I merged some of them

tests/unit/network_graph/test_graph.py
berndbohmeier: done

PR #238, opened 2019-01-07
Feature/optimize least cost path issue 237

schmir:
see https://github.com/trustlines-network/relay/issues/237

### DIFF
--- a/relay/network_graph/alg.py
+++ b/relay/network_graph/alg.py
@@ -79,6 +79,7 @@ def _least_cost_path_helper(
 ):
     graph_adj = graph.adj
 
+    visited_nodes = set()  # set of nodes, where we already found the minimal path
     while queue:
         cost_from_start_to_node, node = heapq.heappop(queue)
         if node in target_nodes:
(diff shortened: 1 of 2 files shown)


relay/network_graph/alg.py
```diff
@@ -79,6 +79,7 @@ def _least_cost_path_helper(
 ):
     graph_adj = graph.adj
 
+    nodes_done = set()  # set of nodes, where we already found the minimal path
```
cducrest: I would rename it visited_nodes since it is the set of nodes the algorithm has already visited

relay/network_graph/alg.py
```diff
@@ -87,7 +88,10 @@ def _least_cost_path_helper(
         if cost_from_start_to_node > least_costs[node]:
```
cducrest: I would guess this line is redundant now

schmir: no, we may still have put an element with non-minimal cost onto the queue and we don't want to look at them again.

PR #247, opened 2019-01-17
Feature/cci run e2e

schmir:
this makes circleci run end2end tests, e.g. https://circleci.com/workflow-run/e3920bcf-a2e8-4449-aa2a-8eac3549f3ca

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -181,6 +181,31 @@ jobs:
             docker tag $LOCAL_IMAGE $DOCKER_REPO:latest
             docker push $DOCKER_REPO:latest
 
+  run-end2end-tests:
+    executor: ubuntu-builder
+    environment:
+      DOCKER_REPO: trustlines/relay
+      LOCAL_IMAGE: relay
+    working_directory: ~
+    steps:
+      - setup_remote_docker
+      - attach_workspace:
+          at: '~'
+      - run:
+          name: Checkout end2end repo
+          command: |
+            git clone https://github.com/trustlines-network/end2end.git
+      - run:
+          name: Load docker image
+          command: |
+            docker load --input ~/images/$LOCAL_IMAGE.tar
+      - run:
+          name: run end2end tests
+          command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -181,6 +181,31 @@ jobs:
...
+          command: |
+            git clone https://github.com/trustlines-network/end2end.git
+      - run:
+          name: Load docker image
+          command: |
+            docker load --input ~/images/$LOCAL_IMAGE.tar
+      - run:
+          name: run end2end tests
+          command: |
+            docker tag $LOCAL_IMAGE $DOCKER_REPO
```
berndbohmeier: I feel like this does more belong to the step above (Load docker image)

schmir: I can't think of a reason, why we should move the tagging to 'load docker image'.  Can you give some other than your feeling?

There are multiple reasons not to do it:
- 'Load docker image' has a counterpart 'Save docker image'. The common expectation would be that after the 'Load docker image' operation we have the same state then before the 'Save docker image' operation. If we tag inside those 'Load docker image' that wouldn't be true anymore.
Or in other words if we look at the steps that are being executed we expect the 'persist to workspace' being cancelled out by 'attach workspace' and 'save docker image' to be cancelled out by 'load docker image'. If we wrote down the steps with those 4 being cancelled out, we would miss tagging the image and test with the  one from dockerhub.

- We already have the 'Load docker image' step inside this configuration end everywhere else where we build docker images. Tagging is not part of it anywhere.

- It's easy to miss the step of tagging the locally build image and you won't notice immediately.  Tagging the image or images before running the script  is a part of running the end2end tests. Everyone running the `run-e2e.sh` script should be aware of it.

berndbohmeier: I dont know for me it is part of loading the image and making it available, but I understand your arguments, then we could also think of it as another step. I definitely fail to see how it is part of running the e2e tests, it is rather a setup to run them right. Maybe if we rename it to running e2e test with local image I am fine. I dont have super strong feelings on this so we can also leave it.

PR #252, opened 2019-01-24
Feature/black

schmir:
[black](https://github.com/ambv/black) is an automatic formatter for python code. We already use it in the py-eth-index and the watch repository.

It does a better job of formatting our source code than as what I am able to do. Therefore I'd like to use it  for all of our python code.

This PR contains the changes to enable it on circle ci and also contains the reformatted source code.

There's wide support for integrating black into your text editor/IDE:
https://black.readthedocs.io/en/stable/editor_integration.html

black calls itself 'The uncompromising Python code formatter'. I think there's only two things you could configure, but I'd like to use the default settings (which is what I used in this PR).

To keep the discussion short on this one, would you be fine with us using black with the default settings as seen in this PR?

If IDE support is a problem I'm willing to help. I can also show you guys how to use the pre-commit hook framework, which you should do even without this PR getting merged. I've added pre-commit files to most of the repos I touched.


cducrest:
I don't really like the rule about closing brackets being dedented and on a new line just to make short diffs. This seems to actually make longer and less readable diffs (see https://github.com/trustlines-network/relay/pull/252/files#diff-edec4b7c2bd8cfc6569aac80adbed914R81).
But I guess if everyone else like it, I can't really oppose.

Can I suggest keeping the char per line limit of 119 instead of the black default of 88 to avoid having too many of these occurrences?

weilbith:
@cducrest _Black_ is about no opinion. It is _Black_'s opinion and that's it. What you are doing locally is your stuff. You can establish a pull hook which formats the code how you like to edit. As long as the repository has the same code format standard, the goal is reached.

berndbohmeier:
@cducrest This makes the diffs shorter if we would add something now, but you are right, that it makes the diffs longer if we go from bellow the line limit to above the line limit. 
I am usually a fan of longer line limits, but because black will only put arguments in there own line if the line exceeds the line length limit, we should keep that limit low, because otherwise stuff like
`function(arg1, arg2, arg3, arg4, arg, arg6, ...)` will stay untouched.

cducrest:
@weilbith Nice, I have to take time to look into reformatting it locally to longer lines and I'll be fine with black I guess.

berndbohmeier:
I feel in the section you mentioned the code became less readable, but I think overall the advantages outweigh the disadvantages.

weilbith:
@cducrest in case you are using only online diffs which GitHub provide: This is only a service they provide as a view to their database of the repository. You can just fetch the changes, do you local formatting again and create a diff view locally. There is no need to be satisfied by GitHub for things you can do locally in a way that support your personal workflow better.

schmir:
To sum up our discussion during the daily: Some folks do have objections against using it, because they think in some cases the readability got worse and they evaluate other tools.
My take on this is that this makes little sense because you will always be able to find cases where 'hand formatted' code may 'look better' than an automatic tool and black looks like it may become the standard python formatting tool.

schmir:
short intro to pre-commit: https://github.com/trustlines-network/project/wiki/pre-commit

schmir:
I've got some more thoughts on this issue. I'm sorry, but it's important enough for me to bring this back to the table.

## on the choice of black vs other formatters

I am following the python ecosystem for some months again. It's pretty clear that black has a lot of momentum now. At least three of our dependencies are already using it (pytest, tox, attrs) and we're using it for the ethindex and watch repositories. 

The raiden folks are also thinking about using it.

I did not run into any problems while using it for the ethindex and watch repos. I've even formatted code for the relay server with black even so I was only committing reformatted parts that I also did change. No one complained about those changes.

Given the above reasons and the fact that we don't have to discuss configuration of the formatter (because black has very little to configure), I fail to see a good reason to look at other formatters at all. 

You will end up with some code anyway, where you might think this would look a bit better if formatted by hand. Others from the team might see this differently. And that's the exact reason why I prefer a tool, which 
cannot be configured.

I can assure you that even if you don't like some parts of the formatting, after adapting to it, you will like the fact that every file looks the same. If you look at our code, there's a still a lot of variance of what it looks like.

I am glad that we don't have a strict style guide and we do not have to complain about formatting issues in our PRs. flake8 will already have complained and using an automatic formatter will bring us one step further in the sense that we do not have to fix those flake8 complains manually.

Anyway, I'm not sure how we ended up refusing this PR, because most people where like "love it"  and come was like ok, we can do it if everyone else likes it. 

Please give it a second thought, thanks.

cducrest:
> I am glad that we don't have a strict style guide and we do not have to complain about formatting issues in our PRs. flake8 will already have complained and using an automatic formatter will bring us one step further in the sense that we do not have to fix those flake8 complains manually.

I thought the reason we would want to use an auto formatter is to avoid discussing style on PR. If we do not discuss style on PR, I do not see the benefit of using an auto formatter.

I fail to see what we gain by avoiding configuring an auto formatter. The time we gain by avoiding discussion is lost in the long run by having less readable code.

After some research and tests, notably reading articles like https://medium.com/3yourmind/auto-formatters-for-python-8925065f9505. I would prefer using yapf with the google style (and possibly more char per lines).

Otherwise, as already said I am fine with using black if everyone loves it. But I do not agree with the arguments given in favour of using it.

schmir:
>I thought the reason we would want to use an auto formatter is to avoid discussing style on PR.

No, that's not the reason. We don't do that at the moment  even though we do format code in an inconsistent style.

At the moment I have to manually please flake8. I want to avoid that. black allows me to do that automatically (and at least to me it looks much better than what we are able to do). I don't have to think about formatting code anymore. In that sense it frees my mind from the mental burden of having to format code manually. 

> I fail to see what we gain by avoiding configuring an auto formatter

We don't need the discussion now or in the future when someone comes along and thinks we should change some rules/configuration. This is something I really want to avoid.

> The time we gain by avoiding discussion is lost in the long run by having less readable code.

I disagree. Having a standard formatter makes all code look the same. You will adapt to it. It's not less readable.  There is value in having all code look the same. To me it looks like the python community as a whole moves into black's direction(if at all,I may live in my own bubble, but as I mentioned three of our dependencies already use it).

It's what the go people do (gofmt), what javasript  is doing (prettier), what elixir is doing (mix format) and what the python people are doing (black). None of the code formatters there allow (much) configuration.

As I said there is value in having all code look the same for a community as a whole. I fail to see a real competitor  to black in the python ecosystem. This is why I consider black the only viable option.

> I would prefer using yapf with the google style 

See, now we're already discussing style.

This is something I woud refuse to do given the reasons above. Another reason is that we already have black in use at some projects.

schmir:
yapf comes with it's own set of problems (non deterministic formatting??), see https://news.ycombinator.com/item?id=17155048 for a discussion.

berndbohmeier:
As we said today, we are fine with using it. Maybe we can check of it makes sense to use a little bit longer line lengths.

schmir:
on black's line length limit: https://black.readthedocs.io/en/stable/the_black_code_style.html#line-length

I've tried it with a limit of 100 and what I can say now, is that I like having longer expressions split into multiple lines instead having them crammed into one line. It requires less brain power to parse expressions, because sub-expressions often start in a new line. Therefore I would prefer to keep the default setting.

The other reasons to prefer the default  are that it requires less effort to configure IDEs and we are already using the defaults in py-eth-index and watch and wouldn't need to change them there and reformat the code.

berndbohmeier:
Its a valid point, I am fine with using it with the default settings and just hope we will not regret it at some point.

weilbith:
I completely support @schmir's both arguments.

schmir:
I've rebased the changes on the current develop branch. 

I've added `pre-commit` to the requirements and added some documentation on how to set it up (which is easy, now that it's installed as part of the requirements).

I've changed comments at two places to enhance the formatting.

@berndbohmeier feel free to take another look at this now.

berndbohmeier:
Can be merged I think

PR #255, opened 2019-01-25
Fix/broken deps

schmir:


berndbohmeier:
Why did the tests not fail because on this?

PR #257, opened 2019-01-25
pass timestamp to get_account_sum functions from TrustlinesRelay

schmir:
get_account_sum otherwise uses the current time and since we call multiple times
into this function, TrustlinesRelay may generate events based on different
timestamps.

While I'm here, also try to be a bit more functional and separate event creation
from publishing the events to the subscribers.

see https://github.com/trustlines-network/relay/issues/222

### DIFF
--- a/relay/relay.py
+++ b/relay/relay.py
@@ -517,12 +517,11 @@ class TrustlinesRelay:
                              balance_update_event.to,
                              balance_update_event.value,
                              balance_update_event.timestamp)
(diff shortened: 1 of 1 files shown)


berndbohmeier:
LGTM
I used this to completely removed time.time() from graph https://github.com/trustlines-network/relay/commit/9597ab865f44b7befd558519c1c2ec7037ffe51f

PR #264, opened 2019-02-07
Feature/eth transaction from meta transaction

cducrest:
Uses functions from a non-released version of tldeploy package.
The CI will fail.

closes https://github.com/trustlines-network/relay/issues/261


tests/chain_integration/test_delegate_transaction.py
```diff
@@ -0,0 +1,91 @@
...
+def delegate_address(web3):
+    return web3.eth.coinbase
+
+
+@pytest.fixture(scope='session')
+def delegate(web3, delegate_address):
+
+    trustlines_relay = TrustlinesRelay()
+    trustlines_relay._load_contracts()
+    identity_contract_abi = trustlines_relay.contracts['Identity']['abi']
```
berndbohmeier: I think this is to big of a dependency, to use the trustlinesRelay object here just to load the contracts. Please just use the `contracts` fixture that can be found in `chain_integration/conftest.py`

PR #265, opened 2019-02-08
Feature/meta transaction rest api 263

schmir:
add the /api/v1/relay-meta-transaction REST API endpoint
    
This adds an endpoint, which is used to relay meta transactions. The
MetaTransaction itself is nested insided a dictionary under the metaTransaction
key. This will allow us to add additional fields.
    
    implements https://github.com/trustlines-network/relay/issues/263


docs/RelayAPI.md
```diff
@@ -766,3 +766,38 @@ curl --header "Content-Type: application/json" \
...
+
+
+The MetaTransaction object must have the following fields:
+
+| Name       | Type    | Description                                                        |
+|------------|---------|--------------------------------------------------------------------|
+| from       | address | address of identity contract                                       |
+| to         | address | the address on which the call of the meta transaction is happening |
+| value      | uint256 | the amount of wei to be sent along from 'from' to 'to'             |
+| data       | bytes   | the data object encoding the function call                         |
```
cducrest: "the data object encoding the function call" ... and the arguments of the call.

tests/unit/api/test_schemas.py
```diff
@@ -0,0 +1,43 @@
...
+    ]
+
+
+@pytest.mark.parametrize("meta_transaction", gen_meta_transactions())
+def test_meta_transaction_schema(meta_transaction):
+    dumped = schemas.MetaTransactionSchema().dump(meta_transaction).data
+    print("DUMPED", dumped)
+
+    loaded = schemas.MetaTransactionSchema().load(dumped).data
+    print("LOADED", loaded)
```
cducrest: remove prints

schmir: no, I know that bernd also always prays to remove those,  but they will be helpful if the tests fail at some point.

PR #270, opened 2019-02-12
add REST endpoint returning information about an identity contract

schmir:
This returns the balance, identity (address) and the lastNonce. The 'create'
endpoint that deploys a new identity contract has been changed to return the
same values (i.e. it returns the additional fields balance and lastNonce, which
should both be zero).

implements https://github.com/trustlines-network/relay/issues/266


relay/api/resources.py
```diff
@@ -354,19 +356,41 @@ def post(self):
         return self.trustlines.node.send_ether(address)
 
 
+def _get_identity_info(identity):
+    try:
+        return {
+            "balance": identity.web3.eth.getBalance(identity.address),
+            "identity": identity.address,
+            "lastNonce": identity.functions.lastNonce().call(),
```
berndbohmeier: I dont think it is a good idea to have the api code have a direct dependency on the function name in the contract. 
I already created the function name for next nonce:
https://github.com/trustlines-network/contracts/blob/develop/py-deploy/tldeploy/identity.py#L183
please add the one for balance there too and then move the code to 
https://github.com/trustlines-network/relay/blob/develop/relay/blockchain/delegate.py

schmir: I've actually looked at identity.py. The problem with the `Identity` class is, that it needs `owner_private_key` as additional argument and Delegator needs `delegator_address` as additional argument. 
I don't have any of those arguments in the api code. 
What I could do now is create a `BaseIdentity` class that provides `get_next_nonce`, `address` and `get_balance` and which would be used as the base class of `Identity`. 
The fact that I need to introduce another class to implement some basic functionality makes me think that we could improve this a bit. Also if we had that `BaseIdentity` wrapper it should probably be used by the `Delegator` class itself. 

My suggestion would be to merge this as is (it doesn't hurt that much,right?), and open an issue about improving `identity.py` in the way I describe above.

> please add the one for balance there too and then move the code to
> https://github.com/trustlines-network/relay/blob/develop/relay/blockchain/delegate.py

I totally do not know what your plan is here and what should be moved. Can you provide some more details?

berndbohmeier: It hurts, because it creates a dependency from the top of our stack(rest API) to the bottom of our stack (smart contract code)
If we change the smart contract this will break, no one will notice because no one will look in the API code and we have no test that will catch it. This is not acceptable.

berndbohmeier:
I can refactor this tomorrow and can show you how I would do it.

schmir:
> I can refactor this tomorrow and can show you how I would do it.

yes, please do that.

PR #275, opened 2019-02-14
Add function for find_path receiver pays for fees.

cducrest:
Modify tests so that the paths tests run with both find_path functions.
Add test for fees of two find_path functions

### DIFF
--- a/relay/network_graph/graph.py
+++ b/relay/network_graph/graph.py
@@ -548,9 +548,17 @@ class CurrencyNetworkGraph(object):
                              'Creditline BA': account.reverse_creditline})
         return output.getvalue()
 
-    def find_path_sender_pays_fees(self, source, target, value=None, max_hops=None, max_fees=None, timestamp=None):
+    def find_transfer_path_sender_pays_fees(
+            self,
+            source,
+            target,
+            value=None,
+            max_hops=None,
+            max_fees=None,
+            timestamp=None,
+    ):
 
-        cost, path = self._find_path(
+        cost, path = self._find_transfer_path(
             source=target,  # we are searching path from target to source, to accumulate fees correctly.
             target=source,
             value=value,
@@ -562,9 +570,17 @@ class CurrencyNetworkGraph(object):
 
         return cost, list(reversed(path))
 
-    def find_path_receiver_pays_fees(self, source, target, value=None, max_hops=None, max_fees=None, timestamp=None):
+    def find_transfer_path_receiver_pays_fees(
+            self,
+            source,
+            target,
+            value=None,
+            max_hops=None,
+            max_fees=None,
+            timestamp=None,
+    ):
 
-        return self._find_path(
+        return self._find_transfer_path(
             source=source,
             target=target,
             value=value,
@@ -574,7 +590,7 @@ class CurrencyNetworkGraph(object):
             cost_accumulator_function=ReceiverPaysCostAccumulatorSnapshot,
         )
 
-    def _find_path(
+    def _find_transfer_path(
             self,
             *,
             source,
@@ -739,7 +755,7 @@ class CurrencyNetworkGraphForTesting(CurrencyNetworkGraph):
 
     def mediated_transfer(self, source, target, value):
         """simulate mediated transfer off chain"""
-        cost, path = self.find_path_sender_pays_fees(source, target, value)
+        cost, path = self.find_transfer_path_sender_pays_fees(source, target, value)
         assert path[0] == source
         assert path[-1] == target
         return self.transfer_path(path, value, cost)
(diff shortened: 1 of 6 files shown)


schmir:
I do think it would make sense to not add a second function and instead add another parameter to `find_path`, which specifies who pays for the fees.
This way you don't have to change any test function.
I've done something similar in an old branch. Please take a look here:
https://github.com/trustlines-network/relay/commit/38f144d1b81f79d11bb87a8e10e9f9e4cdede761#diff-dd5e27e60352c8c9a603b6a45805b32dR20

relay/network_graph/graph.py
```diff
@@ -548,36 +548,66 @@ def dump(self):
...
+            source=source,
+            target=target,
+            value=value,
+            max_hops=max_hops,
+            max_fees=max_fees,
+            timestamp=timestamp,
+            cost_accumulator_function=ReceiverPaysCostAccumulatorSnapshot,
+        )
+
+    def _find_path(
```
berndbohmeier: lets maybe use `_find_transfer_path` to be clear?

berndbohmeier: Then it would be also `find_transfer_path_receiver_pays_fees`

tests/unit/network_graph/test_graph.py
```diff
@@ -88,6 +88,17 @@ def complex_community_with_trustlines_and_fees_configurable_balances(complex_com
     return complex_community_with_trustlines
 
 
+@pytest.fixture(
+    params=[
+        CurrencyNetworkGraph.find_path_sender_pays_fees,
+        CurrencyNetworkGraph.find_path_receiver_pays_fees
+    ]
+)
+def find_path_sender_then_receiver_pays(community_with_trustlines, request):
```
berndbohmeier: I think this is a bad name for a fixture. The idea of a fixture there is to be transparent in my opinion. So for me the name should be something like `find_transfer_path_function` or so.

berndbohmeier:
We could also just have a `payer` variable. `find_transfer_path(..., payer = payer.SENDER)` with possible values (`payer.SENDER` and `payer.RECEIVER`). And let `payer.SENDER` be the default. But I am also fine the current approach with two functions.

relay/network_graph/graph.py
cducrest: Ok, I will change it.

tests/unit/network_graph/test_graph.py
cducrest: I don't understand how `find_transfer_path_function` is transparent.
Then I will use this fixture for finding a path not knowing if it will find a path where the sender pays for fees, the receiver pays for fees, or both.
But I can add `_function` to the name of the fixture since it indeed returns a function and does not find a path.
I called it `parametrised_find_transfer_path_function`. Tell me if it is fine for you.

cducrest:
I talked with Ralf and decided to keep two functions until it is proven to be worse than using a parameter (e.g. if we ever need to add another function where different user pays for the fee). I believe limiting the number of parameters (or set of logically disjoint parameters) enhance readability.

Since there are conflicts with develop due to the black PR, and there is another branch `feature/find-path-receiver-pays-API` based on this one, I would like not to rebase/merge this and resolve the conflicts on the api branch instead.

tests/unit/network_graph/test_graph.py
berndbohmeier: > Then I will use this fixture for finding a path not knowing if it will find a path where the sender pays for fees, the receiver pays for fees, or both.

I assumed that is exactly the goal of that fixture. It will give you a function that finds a transfer path, but you dont specify which one. All tests using that should not rely on weather the sender or receiver pays.

berndbohmeier: If you assume that it contains both functions in that order, than I think you should not group them into a single fixture.

berndbohmeier: Example: You its fine to have a list `fruits = [apple, banana, ...]`
but I think it is wrong to have a list `first_apple_then_banana: [apple, banana]`

cducrest: I don't care about the order, I had it named `find_path_sender_and_receiver_pays`, but I thought one might think both the sender and the receiver pay for the fees (like a shared fee).

However, not all tests using `find_transfer_path_receiver_pays_fees` can use `find_transfer_path_sender_pays_fees` (for example if they test the fees calculations), which is why they cannot use this parametrised fixture and I do not want to name it `find_transfer_path_function`.

If the fixture was named `find_transfer_path_function` I would use it in `test_path_with_fees`, which is wrong since the fee returned by the function would not be the same depending on whether the sender or receiver pays for the fee. (see https://github.com/trustlines-network/relay/pull/275/files#diff-1cc3c064efef43ebd1fda8e62895a381R559)

PR #276, opened 2019-02-18
Remove time.time() from graph

schmir:
I think we forgot to merge this one, so I'm opening a PR for it. 

AFAICT this is the last step necessary to  completely fix https://github.com/trustlines-network/relay/issues/222

@berndbohmeier  Any other reason why we didn't merge this?

--- 
Remove the usage of time.time() completly from graph to make it not
depending in the time. Also remove time.time() from event creation, the
timestamp has now to be provided.
Changed the default parameter of timestamp from None to 0, so that you
can still not provide it if the timestamp is not relevant. To protect
from accidentially not providing the timestamp when it is relevant, I
added a assertion in the relevant calc interest calculation.

### DIFF
--- a/relay/api/messaging/resources.py
+++ b/relay/api/messaging/resources.py
@@ -1,3 +1,5 @@
+import time
+
 from flask_restful import Resource
 from webargs import fields
 from webargs.flaskparser import use_args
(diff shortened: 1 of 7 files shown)


berndbohmeier:
No, I mentioned it here https://github.com/trustlines-network/relay/pull/257 and my plan was that if you like it, you would merge it there. But i guess we forgot.

schmir:
rebased against the current develop branch (which contains the black changes) in branch `fix/removetimefromgraph-black`

schmir:
merged the rebases version as PR #277

PR #280, opened 2019-02-19
Feature/find path receiver pays api

cducrest:
Includes the commits from https://github.com/trustlines-network/relay/pull/275.

The diffs is longer than necessary because I had to rebase on develop after the black commit and run black on multiple commits at the same time.

I used a `sender_pays_fees` bool parameter in the API instead of creating a new API to avoid duplication of multiple functions based on the discussion of previous PR.

Closes https://github.com/trustlines-network/relay/issues/271

### DIFF
--- a/relay/api/resources.py
+++ b/relay/api/resources.py
@@ -33,7 +33,7 @@ from relay.relay import TrustlinesRelay
 from relay.concurrency_utils import TimeoutException
 from relay.logger import get_logger
 
-from relay.network_graph.dijkstra_weighted import PaymentPath
+from relay.network_graph.payment_path import PaymentPath
 
 logger = get_logger("api.resources", logging.DEBUG)
 
@@ -418,6 +418,11 @@ class Path(Resource):
         "maxFees": fields.Int(required=False, missing=None),
         "from": custom_fields.Address(required=True),
         "to": custom_fields.Address(required=True),
+        "fee_payer": fields.Str(
+            required=False,
+            validate=validate.OneOf(["sender", "receiver"]),
+            missing="sender",
+        ),
     }
 
     @use_args(args)
@@ -431,18 +436,35 @@ class Path(Resource):
         value = args["value"]
         max_fees = args["maxFees"]
         max_hops = args["maxHops"]
+        sender_pays_fees = args["fee_payer"] == "sender"
 
-        cost, path = self.trustlines.currency_network_graphs[network_address].find_transfer_path_sender_pays_fees(
-            source=source,
-            target=target,
-            value=value,
-            max_fees=max_fees,
-            max_hops=max_hops,
-            timestamp=timestamp,
-        )
+        if sender_pays_fees:
+            cost, path = self.trustlines.currency_network_graphs[
+                network_address
+            ].find_transfer_path_sender_pays_fees(
+                source=source,
+                target=target,
+                value=value,
+                max_fees=max_fees,
+                max_hops=max_hops,
+                timestamp=timestamp,
+            )
+        else:
+            cost, path = self.trustlines.currency_network_graphs[
+                network_address
+            ].find_transfer_path_receiver_pays_fees(
+                source=source,
+                target=target,
+                value=value,
+                max_fees=max_fees,
+                max_hops=max_hops,
+                timestamp=timestamp,
+            )
 
         return _estimate_gas_for_transfer(
-            self.trustlines, PaymentPath(cost, path, value), network_address
+            self.trustlines,
+            PaymentPath(cost, path, value, sender_pays_fees=sender_pays_fees),
+            network_address,
         )
 
 
--- a/relay/api/resources.py
+++ b/relay/api/resources.py
@@ -432,7 +432,7 @@ class Path(Resource):
         max_fees = args["maxFees"]
         max_hops = args["maxHops"]
 
-        cost, path = self.trustlines.currency_network_graphs[network_address].find_path(
+        cost, path = self.trustlines.currency_network_graphs[network_address].find_path_sender_pays_fees(
             source=source,
             target=target,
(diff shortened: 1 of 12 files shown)


relay/api/resources.py
```diff
@@ -431,18 +432,35 @@ def post(self, args, network_address: str):
...
+                value=value,
+                max_fees=max_fees,
+                max_hops=max_hops,
+                timestamp=timestamp,
+            )
 
         return _estimate_gas_for_transfer(
-            self.trustlines, PaymentPath(cost, path, value), network_address
+            self.trustlines,
+            PaymentPath(cost, path, value, sender_pays_fees=sender_pays_fees),
```
berndbohmeier: What is your reason to add `sender_pays_fees` here? If you want to return in from the api, you are missing the schema.

relay/api/resources.py
```diff
@@ -418,6 +418,7 @@ def __init__(self, trustlines: TrustlinesRelay) -> None:
         "maxFees": fields.Int(required=False, missing=None),
         "from": custom_fields.Address(required=True),
         "to": custom_fields.Address(required=True),
+        "sender_pays_fees": fields.Bool(required=False, missing=True),
```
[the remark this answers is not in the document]
schmir: Right. That's the only complaint I dare to mention. (The other complaint has something to do with naming, but you already discussed that at length yesterday).

relay/api/resources.py
cducrest: Because to estimate the gas necessary for the transaction, you need to know whether you estimate the gas for the function `transfer` or `transferReceiverPays` on the currency network contract (https://github.com/trustlines-network/relay/pull/280/files#diff-fb784d1467ed2b14b3b247b1e09307e3R226).
Or do I misunderstand you?

I do not want to return this value.

relay/api/resources.py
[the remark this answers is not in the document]
cducrest: I changed it to `fee_payer` string

relay/api/resources.py
berndbohmeier: Why do you not want to return that value? I think for clarity it would be good to do so.

berndbohmeier: IYou also need to change the function here https://github.com/trustlines-network/relay/blob/bcaf98519ac599508087bf39fca1a391fe1b816c/relay/api/resources.py#L406. 
It will return the Payment path but always with sender as the payer, which mught be wrong. 
I would say this is partially because the name of this function is bad. We need to give it a better name that describe what it is actually doing: estimating the gas and then returning a path object according to that information.

relay/network_graph/payment_path.py
```diff
@@ -7,4 +7,5 @@ class PaymentPath:
     fee: int
     path: List
     value: int
+    sender_pays_fees: bool = attr.ib(default=True)
```
berndbohmeier: I dont think it is a good idea to have this as default. As you can see this already created the problem that you forgot further up to fill it with the right value.

relay/network_graph/graph.py
```diff
@@ -718,12 +750,14 @@ def close_trustline_path_triangulation(
...
         except nx.NetworkXNoPath:
-            return PaymentPath(fee=0, path=[], value=value)
+            return PaymentPath(fee=0, path=[], value=value, fee_payer=FeePayer.SENDER)
 
         if balance < 0:
             path.reverse()
 
-        return PaymentPath(fee=cost[0], path=path, value=value)
+        return PaymentPath(
+            fee=cost[0], path=path, value=value, fee_payer=FeePayer.SENDER
```
berndbohmeier: This should not always return sender. It should return receiver, if the receiver has to pay for this transfer. You might argue that this is not necessary because sender and receiver are the same person, but if you want to use this to settle and you will just use a normal transfer, you need to know if you use transfer where the sender pays or where the receiver pays, so we should just return here the right value.

relay/network_graph/payment_path.py
```diff
@@ -1,10 +1,17 @@
 from typing import List
 import attr
+from enum import Enum
+
+
+class FeePayer(Enum):
```
berndbohmeier: I like the enum :)

cducrest: I did not changed it. :)

relay/api/resources.py
```diff
@@ -431,20 +442,39 @@ def post(self, args, network_address: str):
...
+            cost, path = self.trustlines.currency_network_graphs[
+                network_address
+            ].find_transfer_path_receiver_pays_fees(
+                source=source,
+                target=target,
+                value=value,
+                max_fees=max_fees,
+                max_hops=max_hops,
+                timestamp=timestamp,
+            )
```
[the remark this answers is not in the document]
cducrest: I changed it.

relay/blockchain/currency_network_proxy.py
```diff
@@ -216,15 +217,21 @@ def estimate_gas_for_transfer(self, sender, receiver, value, max_fee, path):
...
-        ).estimateGas({"from": source})
+
+        if payment_path.fee_payer is FeePayer.SENDER:
+            return self._proxy.functions.transfer(
+                target, payment_path.value, payment_path.fee, payment_path.path[1:]
+            ).estimateGas({"from": source})
+        elif payment_path.fee_payer is FeePayer.RECEIVER:
+            return self._proxy.functions.transferReceiverPays(
+                target, payment_path.value, payment_path.fee, payment_path.path[1:]
+            ).estimateGas({"from": source})
```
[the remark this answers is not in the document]
cducrest: Same as above.

relay/network_graph/graph.py
cducrest: I changed it.

PR #284, opened 2019-03-08
Change the type in the docs from address to string

cducrest:
and uint256 to number.

### DIFF
--- a/docs/RelayAPI.md
+++ b/docs/RelayAPI.md
@@ -795,13 +795,13 @@ The MetaTransaction object must have the following fields:
 
 | Name       | Type    | Description                                                        |
 |------------|---------|--------------------------------------------------------------------|
-| from       | address | address of identity contract                                       |
-| to         | address | the address on which the call of the meta transaction is happening |
-| value      | uint256 | the amount of wei to be sent along from 'from' to 'to'             |
-| data       | bytes   | the data object encoding the function call including arguments     |
-| nonce      | uint256 | nonce used for replay protection                                   |
-| extra_data | bytes   | bytes extra data for backwards compatibility                       |
-| signature  | bytes   | 65 bytes containing concatenated. v,r,s of the signature           |
+| from       | string   | address of identity contract                                       |
+| to         | string   | the address on which the call of the meta transaction is happening |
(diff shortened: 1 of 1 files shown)


docs/RelayAPI.md
```diff
@@ -795,13 +795,13 @@ The MetaTransaction object must have the following fields:
...
-| from       | address | address of identity contract                                       |
-| to         | address | the address on which the call of the meta transaction is happening |
-| value      | uint256 | the amount of wei to be sent along from 'from' to 'to'             |
-| data       | bytes   | the data object encoding the function call including arguments     |
-| nonce      | uint256 | nonce used for replay protection                                   |
-| extra_data | bytes   | bytes extra data for backwards compatibility                       |
-| signature  | bytes   | 65 bytes containing concatenated. v,r,s of the signature           |
+| from       | string   | address of identity contract                                       |
+| to         | string   | the address on which the call of the meta transaction is happening |
+| value      | number   | the amount of wei to be sent along from 'from' to 'to'             |
```
berndbohmeier: From the schema this is a string.

docs/RelayAPI.md
```diff
@@ -795,13 +795,13 @@ The MetaTransaction object must have the following fields:
...
+| data       | string   | the data object encoding the function call including arguments     |
+| nonce      | number   | nonce used for replay protection                                   |
```
berndbohmeier: Also a string.

docs/RelayAPI.md
berndbohmeier: This is because there are only floats in javascript and big numbers can thus not be represented exactly in a number.

berndbohmeier:
I am also fine with explaining it better. At the moment I think that address is not helping at all, because there is no type address in json. You need to provide the address as a string.

cducrest:
I updated the doc to add a column `JSON Type` describing the basic type in the JSON sent, and specifying that the addresses should be prefixed with `"0x"`. I kept the type column for the "type" we have in python.
I modified the doc where I was responsible for it and I felt there was most confusion. I did not modify the whole doc because it takes time and is not part of the sprint.

PR #287, opened 2019-03-15
Add option to set a fixed gas price suggestion

berndbohmeier:
This is necessary because we had the problem that the rpc call
eth_gasPrice is really slow when the last transaction is a while in
the past.

### DIFF
--- a/config.json
+++ b/config.json
@@ -11,5 +11,8 @@
   "enableEtherFaucet": false,
   "enableRelayMetaTransaction": false,
   "enableDeployIdentity": false,
-  "fixedGasPrice": false
+  "gasPriceComputation": {
+    "method": "rpc",
+    "gasPrice": 0
+  }
 }
--- a/config.json
+++ b/config.json
@@ -10,5 +10,6 @@
   "eventQueryTimeout": 20,
   "enableEtherFaucet": false,
   "enableRelayMetaTransaction": false,
-  "enableDeployIdentity": false
+  "enableDeployIdentity": false,
+  "fixedGasPrice": false
 }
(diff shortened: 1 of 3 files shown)


config.json
```diff
@@ -10,5 +10,6 @@
   "eventQueryTimeout": 20,
   "enableEtherFaucet": false,
   "enableRelayMetaTransaction": false,
-  "enableDeployIdentity": false
+  "enableDeployIdentity": false,
+  "fixedGasPrice": false
```
schmir: This makes me think I could change it to `"fixedGasPrice": true`, which will not work and makes me wonder where I set the gas price. I would prefer the use of negative values here (which would mean ask via rpc) and rename it to "gasPrice".
Or something like
```
"gasPriceComputation": "rpc"   # or "fixed",
"gasPrice": 100
```

schmir: (or not set the field at all, if you don't want a fixed gas price)

berndbohmeier: not set the value at all will also work

berndbohmeier: I think allowing a negative number to disable it, is similar to what we have now, but less explicit. I dont like if special numbers are treated special if you can solve it differently.

berndbohmeier: I can change it to:
```
"gasPriceComputation": {
"method": "rpc",
"gasPrice": 100
}
```
But I think this is less self documenting, because how should I now know that the other option is named "fixed"

berndbohmeier: Now I think someone would try true, this would fail with an error, and then she would probably just set it to a number.

schmir: The gasPriceComputation map is fine.

PR #298, opened 2019-07-08
Remove mtime update on TrustlineUpdate event

berndbohmeier:


berndbohmeier:
I was not sure if we might need it again in the future, but I can remove it if you want.

berndbohmeier:
it would then have to change also the calling code.

cducrest:
There is only one places that calls it. I believe it makes sense to remove it (and it is easy to put back when we need it).

berndbohmeier:
Thanks

Issue #299, opened 2019-07-19
Relay information of extraData on receiving transfer event


cducrest:
Update relay server to publish the extraData information within the new `Transfer` event.

PR #300, opened 2019-07-22
Feature/add extra data to transfers

cducrest:
Need to release a new version of contracts using the extra data for transfers before seeing if this works with the end2end tests.
See: https://github.com/trustlines-protocol/contracts/pull/208
closes: https://github.com/trustlines-protocol/relay/issues/299

I am not sure about the `HexEncodedBytes` in the schema for the API, but errors should be seen in the e2e tests.

We have an error when using extra data that are not a multiple of 32 bytes. The error is thrown by `eth_abi/decoding.py`: `Tried to read 32 bytes.  Only got 4 bytes`. I had the test for extra data in events working by removing this error throw (adding `return data[:data_length]` before the checks for data length https://github.com/ethereum/eth-abi/blob/master/eth_abi/decoding.py#L517).


relay/blockchain/currency_network_events.py
```diff
@@ -22,7 +21,12 @@ def value(self):
 
 
 class TransferEvent(ValueEvent):
-    pass
+    @property
+    def extra_data(self):
+        extra_data = self._web3_event.get("args").get("_extraData")
+        if extra_data:
+            return f"0x{extra_data.hex()}"
```
berndbohmeier: Use `HexBytes` internally

relay/api/resources.py
```diff
@@ -396,11 +396,16 @@ def get(self, identity_address: str):
 
 
 def _fill_estimated_gas_in_payment_path(
-    trustlines: TrustlinesRelay, payment_path: PaymentPath, network_address: str
+    trustlines: TrustlinesRelay,
+    payment_path: PaymentPath,
+    network_address: str,
+    extra_data="0x",
```
berndbohmeier: Should be `HexBytes`

relay/api/resources.py
```diff
@@ -429,6 +434,7 @@ def __init__(self, trustlines: TrustlinesRelay) -> None:
             validate=validate.OneOf([fee_payer.value for fee_payer in FeePayer]),
             missing="sender",
         ),
+        "extraData": custom_fields.HexEncodedBytes(required=False, missing="0x"),
```
berndbohmeier: missing should be of type `HexBytes`

relay/blockchain/currency_network_proxy.py
```diff
@@ -212,12 +212,14 @@ def get_all_network_events(
         results = concurrency_utils.joinall(queries, timeout=timeout)
         return sorted_events(list(itertools.chain.from_iterable(results)))
 
-    def estimate_gas_for_transfer(self, sender, receiver, value, max_fee, path):
+    def estimate_gas_for_transfer(
+        self, sender, receiver, value, max_fee, path, extra_data="0x"
```
berndbohmeier: extra_data should have type `HexBytes`

relay/blockchain/currency_network_proxy.py
```diff
@@ -212,12 +212,14 @@ def get_all_network_events(
...
+    ):
         return self._proxy.functions.transfer(
-            receiver, value, max_fee, path
+            receiver, value, max_fee, path, extra_data
         ).estimateGas({"from": sender})
 
-    def estimate_gas_for_payment_path(self, payment_path: PaymentPath):
+    def estimate_gas_for_payment_path(self, payment_path: PaymentPath, extra_data="0x"):
```
berndbohmeier: Same

tests/conftest.py
```diff
@@ -59,6 +60,15 @@ def test_account():
     )
 
 
+@pytest.fixture(scope="session")
+def test_extra_data():
+    # Currently the tests are only working with extra_data the length of a multiple of 32 bytes
```
berndbohmeier: This can be removed now?

tests/conftest.py
```diff
@@ -59,6 +60,15 @@ def test_account():
...
+    # It is suspected to be a bug within the eth_abi decoding
+    return hexbytes.HexBytes(
+        "0x1234567812345678123456781234567812345678123456781234567812345678"
```
berndbohmeier: This can be set to something != len 32?

tests/conftest.py
cducrest: Yes, planned to do that

tests/conftest.py
cducrest: Yes, planned to do that

PR #301, opened 2019-08-05
Update dependencies 2019 08 05

schmir-at-bb:
This doesn't upgrade the dependencies that are only used in the relay
server. I fear I might break some stuff. We need to update those later (mostly the firebase stuff and marshmallow).

The shared dependencies are in sync with all other repos.


berndbohmeier:
Needs to be merged/rebased. I would do it, but it is not so easy to do because you use your own branch.

schmir-at-bb:
The problem with the end2end tests can also be observed on the current head of the master branch. See https://circleci.com/workflow-run/471614c0-f45f-41f3-9098-f86aa38ab532
I will wait for that one to be sorted out. This PR already contains a fix for the mypy issue that's also existent on master.

berndbohmeier:
Merging this, because we need the fix.

PR #306, opened 2019-08-09
Feature/update marshmallow

cducrest:
Updates marshmallow to 3.0.0rc9 and related dependencies.
Based on the dates of the released versions, I decided to go to v3 even though it is an `rc` version and picked the most recent one: https://github.com/marshmallow-code/marshmallow/releases
I have `3.0.0b7` in requirements because as far as I can tell I am not using any features from anything above it.

Still need to run the e2e tests and most likely modify the `missing` value for the api.


cducrest:
It seems to me that when I run the e2e test from this pr: https://github.com/trustlines-protocol/clientlib/pull/239
with or without this commit: https://github.com/trustlines-protocol/relay/commit/68b670dd5dc1db892bd0c347ed56bd0a26cb1047
on this branch, the result is the same. But I may be tired.

berndbohmeier:
I would have expected that we also have to change the way we specify the default value. But it seems to work and we have a test case now for it, so I am gonna merge it.

Issue #310, opened 2019-08-26
Update graph vision for frozen trustlines


weilbith:
- react from the currency network events
- make sure path finding does not use frozen trustlines

PR #312, opened 2019-08-29
Feature/account for frozen trustlines

cducrest:
Closes https://github.com/trustlines-protocol/relay/issues/310

Use unreleased version of contracts with https://github.com/trustlines-protocol/contracts/pull/233 so obviously CI is failing.


relay/blockchain/currency_network_events.py
```diff
@@ -55,6 +55,10 @@ def interest_rate_given(self):
     def interest_rate_received(self):
         return self._web3_event.get("args").get("_interestRateReceived", 0)
 
+    @property
+    def is_frozen(self):
+        return self._web3_event.get("args").get("_isFrozen", False)
```
schmir-at-bb: Why would this value not be set in 'args'? I think you already require the new version, so my guess is '_isFrozen` will be there. If that's the case, please remove the 'False'.

cducrest:
Merged with other PR

PR #324, opened 2019-09-24
Do not keep the session object alive

schmir-at-bb:
The ClientTokenDB class could not be used from multiple greenlets
previously. It resulted in the following errors being thrown:

,----
| <class 'sqlalchemy.exc.ProgrammingError'> (psycopg2.ProgrammingError)
| execute cannot be used while an asynchronous query is underway
`----

We now create a sesssion object for each method and make sure it's
lifetime is being handled correctly with a contextmanager.

Also we have seen the following errors on the development server,
see https://github.com/trustlines-network/deployment-config/issues/40:

,----
| sqlalchemy.exc.InvalidRequestError: This Session's transaction has
| been rolled back due to a previous exception during flush. To begin a
| new transaction with this Session, first issue
| Session.rollback(). Original exception
| was: (psycopg2.OperationalError) SSL connection has been closed
| unexpectedly
`----

These should also be fixed now.

### DIFF
--- a/relay/pushservice/client_token_db.py
+++ b/relay/pushservice/client_token_db.py
@@ -5,6 +5,7 @@ from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy import Column, String
 from sqlalchemy.orm import sessionmaker
 from sqlalchemy.exc import IntegrityError
+from contextlib import contextmanager
 
 Base = declarative_base()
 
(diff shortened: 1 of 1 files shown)


relay/pushservice/client_token_db.py
```diff
@@ -28,23 +29,39 @@ class ClientTokenAlreadyExistsException(ClientTokenDBException):
...
         Base.metadata.create_all(engine)
-        self.session = Session()
+
+    @contextmanager
+    def session(self):
+        """Provide a transactional scope around a series of operations."""
+        session = self._make_session()
+        try:
+            yield session
+            session.commit()
```
berndbohmeier: This always does a session commit, we use it later even if we only do a query, does that make sense?

relay/pushservice/client_token_db.py
```diff
@@ -55,20 +72,20 @@ def add_client_token(self, user_address: str, client_token: str) -> None:
...
-        except IntegrityError as e:
-            self.session.rollback()
-            raise ClientTokenAlreadyExistsException from e
+        with self.session() as session:
+            tokenMappingOrm = TokenMappingORM(
+                user_address=user_address, client_token=client_token
+            )
+            try:
+                session.add(tokenMappingOrm)
+                session.commit()
```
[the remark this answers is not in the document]
schmir-at-bb: This is needed for the next line, where we catch the IntegrityError. Otherwise the error would be raised outside of the try block.

relay/pushservice/client_token_db.py
schmir-at-bb: I actually copied that from the sqlalchemy documentation. We could also remove it and call commit inside the actual methods. I don't think there's a problem with calling commit even if we didn't write to the database.

From what I can tell this will also close the transaction that SQLAlchemy opened for us while querying the database.So, I think this something that we actually should do (doesn't matter in our case, because we close the session later on anyway).

PR #335, opened 2019-10-09
Upgrade contracts version to 0.9.3 to be able to deploy two identitie…

cducrest:
…s in the same block


cducrest:
It seems the end2end tests are flaky because of this line: https://github.com/trustlines-protocol/clientlib/blob/master/tests/e2e/Identity.test.ts#L180

This PR will fix it: https://github.com/trustlines-protocol/clientlib/pull/256. So I am merging this.

PR #336, opened 2019-10-10
[WIP] Fix getting a path from pending block

cducrest:
Add a failing test for using a path that is only made available by a pending block
Getting the path works, but using the path (shown by gas estimate) does not.

I struggle to find a way to fix this.

It seems you cannot give a block identifier to `estimateGas` on a contract function ([docs](https://web3py.readthedocs.io/en/stable/contracts.html#web3.contract.ContractFunction.estimateGas)).

So I go around by building the function without estimating the gas (`transaction = self._proxy.functions.transfer(receiver, value, max_fee, path, extra_data).buildTransaction({"from": sender, "gas": 0})` and then attempt to estimate the gas with `web3.eth.estimateGas()` ([docs](https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.estimateGas)).

However this does not work in the test and throw error: `TypeError: estimate_gas() takes 2 positional arguments but 3 were given`. It seems to me that at least `eth_tester` cannot estimate gas on the pending block: https://github.com/ethereum/eth-tester/blob/master/eth_tester/main.py#L431

Seems like PY-EVM backend can also not do that: https://github.com/ethereum/eth-tester/blob/master/eth_tester/backends/pyevm/main.py#L466

### DIFF
--- a/relay/api/resources.py
+++ b/relay/api/resources.py
@@ -423,20 +423,9 @@ def _fill_estimated_gas_in_payment_path(
     extra_data: hexbytes.HexBytes = hexbytes.HexBytes(b""),
 ) -> PaymentPath:
     proxy = trustlines.currency_network_proxies[network_address]
-    try:
-        payment_path.estimated_gas = proxy.estimate_gas_for_payment_path(
-            payment_path, extra_data
-        )
-    except ValueError:
-        # should mean out of gas, so path was not right.
-        return PaymentPath(
-            fee=0,
-            path=[],
-            value=payment_path.value,
-            estimated_gas=0,
-            fee_payer=payment_path.fee_payer,
-        )
-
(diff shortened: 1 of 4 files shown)


cducrest:
Interestingly enough, the e2e tests are passing. It's just eth-tester being unable to take the block_identifier argument on estimate-gas.

berndbohmeier:
The main issue was in my opinion that our gas estimation was wrong. So maybe if gas estimation fails for some reason we should maybe not return No path, but maybe fall back to a different gas estimation, or only set gas estimation to a value that indicates we failed to provide a good gas estimation. 

If it works for the end2end test, I would remove the test here and use it.

cducrest:
Problem alleviated in a different way in a different PR: https://github.com/trustlines-protocol/relay/pull/340

Issue #350, opened 2019-11-18
Endpoint to query for fees for a meta-tx


cducrest:
The endpoint receives a meta-tx, answers with fees it would want for this meta-tx

Issue #351, opened 2019-11-18
Restrict delegation to meta-tx with certain fees


cducrest:
Only send meta-tx that matches a fee the relay would pay.

PR #353, opened 2019-11-20
Feature/fees for meta tx

cducrest:
closes https://github.com/trustlines-protocol/relay/issues/350

### DIFF
--- a/docs/RelayAPI.md
+++ b/docs/RelayAPI.md
@@ -846,11 +846,8 @@ The MetaTransaction object must have the following fields:
 | to                    | address  | string - hex-encoded prefixed with "0x"   | the address on which the call of the meta transaction is happening |
 | value                 |BigInteger| string                                    | the amount of wei to be sent along from 'from' to 'to'             |
 | data                  | bytes    | string - hex-encoded prefixed with "0x"   | the data object encoding the function call including arguments     |
-| delegationFees        |BigInteger| string                                    | the fees the delegate will receive for the meta transaction        |
-| currencyNetworkOfFees | address  | string - hex-encoded prefixed with "0x"   | the currency network used to pay the fees of the meta transaction  |
 | nonce                 | int      | number                                    | nonce used for replay protection                                   |
 | extraData             | bytes    | string - hex-encoded prefixed with "0x"   | bytes extra data for backwards compatibility                       |
-| signature             | bytes    | string - hex-encoded prefixed with "0x"   | 65 bytes containing concatenated. v,r,s of the signature           |
 
 #### Example Request
 ```bash
--- a/docs/RelayAPI.md
+++ b/docs/RelayAPI.md
@@ -856,7 +856,7 @@ The MetaTransaction object must have the following fields:
 ```bash
 curl --header "Content-Type: application/json" \
   --request POST \
-  --data '{"metaTransaction": {"value": "0", "to": "0x51a240271AB8AB9f9a21C82d9a85396b704E164d", "nonce": "1", "data": "0x46432830000000000000000000000000000000000000000000000000000000000000000a", "from": "0xF2E246BB76DF876Cef8b38ae84130F4F55De395b", "delegationFees": 1, "currencyNetworkOfFees": "0x51a240271AB8AB9f9a21C82d9a85396b704E164d", signature": "0x6d2fe56ef6648cb3f0398966ad3b05d891cde786d8074bdac15bcb92ebfa7222489b8eb6ed87165feeede19b031bb69e12036a5fa13b3a46ad0c2c19d051ea9101", "extraData": "0x"}}' \
+  --data '{"metaTransaction": {"from": "0xF2E246BB76DF876Cef8b38ae84130F4F55De395b", "to": "0x51a240271AB8AB9f9a21C82d9a85396b704E164d", "value": "0", "data": "0x46432830000000000000000000000000000000000000000000000000000000000000000a", "delegationFees": 1, "currencyNetworkOfFees": "0x51a240271AB8AB9f9a21C82d9a85396b704E164d", "nonce": "1", "extraData": "0x", signature": "0x6d2fe56ef6648cb3f0398966ad3b05d891cde786d8074bdac15bcb92ebfa7222489b8eb6ed87165feeede19b031bb69e12036a5fa13b3a46ad0c2c19d051ea9101"}}' \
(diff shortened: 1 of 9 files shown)


docs/RelayAPI.md
```diff
@@ -825,6 +826,56 @@ curl --header "Content-Type: application/json" \
...
+
+The MetaTransaction object must have the following fields:
+
+| Name                  | Type     | JSON Type                                 | Description |
+|-----------------------|--------- |-------------------------------------------|-------------|
+| from                  | address  | string - hex-encoded prefixed with "0x"   | address of identity contract |
+| to                    | address  | string - hex-encoded prefixed with "0x"   | the address on which the call of the meta transaction is happening |
+| value                 |BigInteger| string                                    | the amount of wei to be sent along from 'from' to 'to'             |
+| data                  | bytes    | string - hex-encoded prefixed with "0x"   | the data object encoding the function call including arguments     |
+| delegationFees        |BigInteger| string                                    | the fees the delegate will receive for the meta transaction        |
```
weilbith: Why do we send those two values if they do not matter? As I understood it we ask the delegate where and how much he would take to relay the transaction. Is it ever planned to include some logic that evaluates the "suggestion" of the user?

relay/blockchain/delegate.py
```diff
@@ -6,6 +8,10 @@
     deploy_proxied_identity,
 )
 
+META_TRANSACTION_FEE_VALUE = 1
```
weilbith: Am I correct that this fixed value solution becomes them adjusted by #351 ?

cducrest:
> I'm new to this repo, but when do you theoretically add the test for such a new endpoint?

For endpoint we usually rely on end2end tests: https://github.com/trustlines-protocol/clientlib/pull/269/files?file-filters%5B%5D=#diff-678fd2890eee49f15adad98fe3d457b0R228

docs/RelayAPI.md
```diff
@@ -825,6 +826,56 @@ curl --header "Content-Type: application/json" \
...
+|-----------------|------------------------|----------|------------------------|
+| metaTransaction | object                 | YES      | MetaTransaction object |
+
+
+The MetaTransaction object must have the following fields:
+
+| Name                  | Type     | JSON Type                                 | Description |
+|-----------------------|--------- |-------------------------------------------|-------------|
+| from                  | address  | string - hex-encoded prefixed with "0x"   | address of identity contract |
+| to                    | address  | string - hex-encoded prefixed with "0x"   | the address on which the call of the meta transaction is happening |
```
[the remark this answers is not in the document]
cducrest: I cannot think of a more concise explanation. This doc is not supposed to explain what a meta-transaction is, but how to use the endpoint.

relay/api/app.py
```diff
@@ -116,6 +117,7 @@ def add_resource(resource, url):
 
     if trustlines.enable_relay_meta_transaction:
         add_resource(RelayMetaTransaction, "/relay-meta-transaction")
+        add_resource(MetaTransactionFees, "/meta-transaction-fees")
```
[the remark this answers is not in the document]
cducrest: That would be a breaking change, I do not think it is worth it. (The new version of the relay would not work with the old version of the clientlib because the endpoint changed url)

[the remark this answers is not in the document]
weilbith: True. So it is already to late... I guess it is okay since it will not grow further.

docs/RelayAPI.md
cducrest: It could be that the user wants to suggest some currency network for the fees (based on current stupid implementation, he actually should). For the initial delegationFees I cannot think of a reason to have them for now but if the user does not provide them, a default of `0` will be used: https://github.com/trustlines-protocol/relay/pull/353/files?file-filters%5B%5D=#diff-afdabd145e6edf8eefd3fc20ff77ce93R40

relay/blockchain/delegate.py
cducrest: No, the current user story suggests "For now, we use a flat-flat fee (any meta transaction will be relayed for x - where is can be set)". I would not know right now how to use a smarter value than this.

cducrest: I could make this value adjustable via the config of the relay server but I do not believe it makes much more sense.

docs/RelayAPI.md
cducrest: (There is also a default for currency network of fees)

weilbith: Okay, can we mark them as optional and tell the default?

relay/blockchain/delegate.py
weilbith: So what will #351 be then about? Check if the send fee amount is <= this variable?

cducrest: >=, but yes.

weilbith: Haha, sure. Thanks for your help.

docs/RelayAPI.md
cducrest: You mean in the docs?

weilbith: Yes. I would like to know that I must come with a suggestion and still can get the delegates preferred fee options.

docs/RelayAPI.md
```diff
@@ -825,6 +826,53 @@ curl --header "Content-Type: application/json" \
...
+  --data '{"metaTransaction": {"from": "0xF2E246BB76DF876Cef8b38ae84130F4F55De395b", "to": "0x51a240271AB8AB9f9a21C82d9a85396b704E164d", "value": "0", "data": "0x46432830000000000000000000000000000000000000000000000000000000000000000a", "delegationFees": 1, "currencyNetworkOfFees": "0x51a240271AB8AB9f9a21C82d9a85396b704E164d", "nonce": "1", "extraData": "0x", signature": "0x6d2fe56ef6648cb3f0398966ad3b05d891cde786d8074bdac15bcb92ebfa7222489b8eb6ed87165feeede19b031bb69e12036a5fa13b3a46ad0c2c19d051ea9101"}}' \
+  https://relay0.testnet.trustlines.network/api/v1/meta-transaction-fees
+```
+
+#### Response
+|Attribute|Type|Description|
+|---------|----|-----------|
+|delegationFees|string|Fees to be paid for the meta transaction|
+|currencyNetworkOfFees|string|the currency network used to pay the fees of the meta transaction|
+
```
schmir-at-bb: Can we return a list of those objects please? The delegate might accept fees payed in multiple currency networks, not only in one network.

PR #354, opened 2019-11-26
Add fee validation to send signed meta transaction

cducrest:
closes https://github.com/trustlines-protocol/relay/issues/351

### DIFF
--- a/relay/blockchain/delegate.py
+++ b/relay/blockchain/delegate.py
@@ -75,6 +75,8 @@ class Delegate:
 
     def validate_meta_transaction_fees(self, meta_transaction: MetaTransaction):
         fees_estimations = self.calculate_fees_for_meta_transaction(meta_transaction)
+        if not fees_estimations:
+            return
         for fees_estimation in fees_estimations:
             if fees_estimation.value == 0:
                 return
--- a/relay/blockchain/delegate.py
+++ b/relay/blockchain/delegate.py
@@ -1,4 +1,5 @@
 from collections import namedtuple
+from typing import List
 
 from deploy_tools.deploy import TransactionFailed
 from tldeploy.identity import (
@@ -29,6 +30,7 @@ class Delegate:
         self.delegation_fees = delegation_fees
 
     def send_signed_meta_transaction(self, signed_meta_transaction: MetaTransaction):
+        self.validate_meta_transaction_fees(signed_meta_transaction)
         try:
             valid = self.delegate.validate_meta_transaction(signed_meta_transaction)
         except UnexpectedIdentityContractException as e:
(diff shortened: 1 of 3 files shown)


relay/blockchain/delegate.py
```diff
@@ -64,7 +65,9 @@ def calc_next_nonce(self, identity_address: str):
         except UnexpectedIdentityContractException:
             raise InvalidIdentityContractException
 
-    def calculate_fees_for_meta_transaction(self, meta_transaction: MetaTransaction):
+    def calculate_fees_for_meta_transaction(
+        self, meta_transaction: MetaTransaction
+    ) -> DelegationFees:
```
berndbohmeier: Something seems wrong here with the type hint. 
I wonder why this is not caught by mypy.

relay/blockchain/delegate.py
```diff
@@ -64,7 +65,9 @@ def calc_next_nonce(self, identity_address: str):
...
         try:
             valid = self.delegate.validate_nonce(meta_transaction)
```
berndbohmeier: why do we validate the nonce here?

cducrest: That's the only thing we can validate and it should allow failing earlier if the nonce is wrong.

relay/blockchain/delegate.py
cducrest: If I run `mypy --ignore-missing-imports relay tests` it does not raise any error. But my IDE complains on the line `return self.delegation_fees` by saying "Expected type 'DelegationFees', got 'List[DelegationFees]' instead "

cducrest: It also does not complain when I run `mypy --ignore-missing-imports relay/blockchain/delegate.py `

berndbohmeier: mmh, sad, we could maybe see if we can somehow upgrade or so, to see if that fixes it. 
I'll open an issue.

relay/blockchain/delegate.py
berndbohmeier: yes, I think it is just a weird place to do that.

tests/chain_integration/test_delegate.py
```diff
@@ -255,3 +258,13 @@ def test_delegated_transaction_invalid_identity_contract(
...
+
+
+def test_meta_transaction_fees_valid(delegate, meta_transaction):
+    """
+    Check that no exception is raised when validating a valid meta_transaction
+    """
+    delegation_fees = delegate.calculate_fees_for_meta_transaction(meta_transaction)
+    meta_transaction.fees = delegation_fees.value
+    meta_transaction.currency_network_of_fees = delegation_fees.currency_network
+    delegate.validate_meta_transaction_fees(meta_transaction)
```
berndbohmeier: Where is the test here?

berndbohmeier: nvm, it is said in the comment.

PR #355, opened 2019-11-26
Fix validation of signature in schema so that we can send a meta-tx w…

cducrest:
…ith empty signature to query for fees

### DIFF
--- a/relay/api/schemas.py
+++ b/relay/api/schemas.py
@@ -17,12 +17,12 @@ class MetaTransactionSchema(Schema):
     def _validate(self, data):
         nonce = data["nonce"]
         fees = data["fees"]
-        signature = data.get("signature", "")
+        signature = data["signature"]
         if not 0 <= nonce < 2 ** 256:
             raise ValidationError(f"nonce={nonce} is out of bounds")
         if not 0 <= fees < 2 ** 64:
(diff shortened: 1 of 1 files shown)


relay/api/schemas.py
```diff
@@ -17,12 +17,12 @@ class Meta:
...
         nonce = data["nonce"]
         fees = data["fees"]
-        signature = data.get("signature", "")
+        signature = data["signature"]
         if not 0 <= nonce < 2 ** 256:
             raise ValidationError(f"nonce={nonce} is out of bounds")
         if not 0 <= fees < 2 ** 64:
             raise ValidationError(f"delegationFees={fees} is out of bounds")
-        if len(signature) != 65 and signature != "":
+        if len(signature) != 65 and len(signature) != 2:
```
schmir-at-bb: if len(signature) != 2 is the same as signature != "0x" we should use the latter. If it's something different, you should explain it.

cducrest:
> if the len is 65 bytes, then the signature isn't even a string, right? then it's a bytes object?

I can make it a Hexbytes if you like

schmir-at-bb:
Thanks, that makes much more sense. Sorry, that I didn't spot it on the first look at it.

PR #356, opened 2019-11-29
Change default currency network of fees for delegation to empty

cducrest:
The clientlib will ask for the currency network of fees and
use it for its meta transaction so it should pass the schema
validation which expect an address / missing


berndbohmeier:
Looks ok to me, but apperently the config is at best misleading, so I opened an issue #357

Issue #357, opened 2019-12-02
Change config of delegate fees


berndbohmeier:
Currently the config takes two values, `value` and `currencyNetwork`. Both are written as single noun, but apparently take a list. This is not intuitive and I dont know how it should work. 
Probably the proper solution is to take a list of pairs of (currencyNetwork, value)

schmir-at-bb:
The way to do this is to specify multiple sections:
```
[[relay.delegationFees]]
value = 2
currencyNetwork = "0x002..."

[[relay.delegationFees]]
value = 1
currencyNetwork = "0x0001...."
```
This gives the following data structure:
```
In [9]: d["relay"]["delegationFees"]                                            
Out[9]: 
[{'value': 2, 'currencyNetwork': '0x002...'},
 {'value': 1, 'currencyNetwork': '0x0001....'}]
```
(This is what I meant with https://github.com/trustlines-protocol/relay/pull/353#pullrequestreview-322141476)

PR #358, opened 2019-12-06
Declare delegation fees as TOML array of tables in config

schmir-at-bb:
That's the natural thing to do. We even did that conversion
internally before.

See https://github.com/trustlines-protocol/relay/issues/357


schmir-at-bb:
Please don't hit the merge button on that one yet.

Issue #370, opened 2020-01-10
Function to calculate list of paid interests


cducrest:
Write a function that get a list of paid / received interests with timestamps for a user / a trustline of a user from events.

PR #371, opened 2020-01-13
Feature/list of interests

cducrest:
Add function to get events of a certain type for a trustline
Add function to get list of interests within time window for a trustline

Please look carefully at the requests made in https://github.com/trustlines-protocol/relay/pull/371/files#diff-6caecb6c151ad021cb4ee3f86396524dR238 as they will only be tested via end2end tests (so are untested now).

Closes: https://github.com/trustlines-protocol/relay/issues/370
inspired by: https://github.com/trustlines-protocol/contracts/blob/master/tests/test_information_from_events.py


src/relay/blockchain/events_informations.py
```diff
@@ -0,0 +1,133 @@
...
+from typing import List
+
+from relay.blockchain.currency_network_events import (
+    BalanceUpdateEvent,
+    BalanceUpdateEventType,
+    TrustlineUpdateEventType,
+)
+from relay.network_graph.interests import calculate_interests
+
+InterestAccrued = namedtuple("InterestAccrued", ["value", "interest_rate", "timestamp"])
```
schmir-at-bb: If there is no good reason to use named tuples, please consider using the attrs module. The attrs documentation lists some reasons why you should use that: https://www.attrs.org/en/stable/why.html#namedtuples

src/relay/blockchain/events_informations.py
```diff
@@ -0,0 +1,133 @@
...
+        currency_network_address, user, counterparty, event_name=BalanceUpdateEventType
+    )
+
+    timestamps = [
+        balance_update_event.timestamp for balance_update_event in balance_update_events
+    ]
+
+    balances = get_sorted_balances_from_update_events(balance_update_events)
+    interest_rates = [
+        get_interests_rates_of_trustline_for_user_before_timestamp(
```
schmir-at-bb: The call to `get_interests_rates_of_trustline_for_user_before_timestamp` will ask for the same TrustlineUpdateEvents for each balance update event that you have. That slows down the computation and makes the code harder to understand.
Please write this in a way that you have a function (or maybe even better a class) that does the computation on pure data without making any network or database calls. From what I can tell you only need to provide the BalanceUpdate and TrustlinesUpdate events.

src/relay/blockchain/events_informations.py
```diff
@@ -0,0 +1,133 @@
...
+        InterestAccrued(
+            calculate_interests(balance, interest_rates, post_time - pre_time),
+            interest_rates,
+            post_time,
+        )
+        for (balance, interest_rates, pre_time, post_time) in zip(
+            balances[:-1], interest_rates, timestamps[:-1], timestamps[1:]
+        )
+    ]
+
```
schmir-at-bb: This function is written using multiple list comprehensions, which makes it hard to understand.
From what I can tell you could write this as a single for loop:
```
for (pre_balance_event, post_balance_event) in zip(balance_update_events[:-1], balance_update_events[1:]):
   balance = balance_viewed_from_user(...)
   interest = ...
   retval.append(InterestAccrued(...))
```

src/relay/blockchain/events_informations.py
```diff
@@ -0,0 +1,133 @@
...
+    """Get all balances of a trustline in sorted order from the view of user"""
+
+    def balance_viewed_from_user(balance_update_event):
+        if balance_update_event.direction == DIRECTION_SENT:
+            return balance_update_event.value
+        elif balance_update_event.direction == DIRECTION_RECEIVED:
+            return -balance_update_event.value
+        else:
+            RuntimeError("Unexpected balance update event")
+
```
schmir-at-bb: `raise RuntimeError(...)`

src/relay/blockchain/events_informations.py
cducrest: I don't think the reasons listed are good enough to make me use attrs. 

> "If you want a tuple with names, by all means: go for a namedtuple." 

Yes.

But I'm biased by the fact that when we use attrs my editor complains that every use of it is unexpected.

src/relay/blockchain/events_informations.py
cducrest: I changed it as you suggested.

src/relay/blockchain/events_informations.py
cducrest: I changed it for a for loop as you suggest.

src/relay/blockchain/events_informations.py
cducrest: done.

src/relay/blockchain/events_informations.py
schmir-at-bb: no, you do not iterate over consecutive pairs of BalanceUpdateEvents as I suggested. If you do that, you can easily get rid of the assignments to balances and timestamps. 

Please use 
```
for (pre_balance_event, post_balance_event) in zip(balance_update_events, balance_update_events[1:]):
  ...
```
(i.e. the zip expression can also be simplified. It's using only two arguments and only one slice operation)

schmir-at-bb: Strictly speaking you could even introduce a function to iterate over consecutive pairs. But, we have toolz as an indirect dependency, so please use sliding window: https://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.sliding_window

src/relay/blockchain/events_informations.py
schmir-at-bb: pycharm should support it: https://blog.jetbrains.com/pycharm/2018/06/pycharm-2018-2-eap-2-2/ 

If that is your only reason to keep using the namedtuple, please change it.

src/relay/blockchain/events_informations.py
schmir-at-bb: > I changed it to get all necessary events at the beginning of the function get_list_of_paid_interests_for_trustline and pass the events to utility functions. 

good, but there's now a natural point to extract the computation that happens afterwards into it's own function function. So, please split the function. 

What happens before the computation could be made a method in the events_proxy class. But I'm also fine with keeping it as a function.

> We could further optimize get_interests_rates_of_trustline_for_user_before_timestamp

I think that's not necessary. Let's keep this simple.

src/relay/blockchain/events_informations.py
```diff
@@ -1,20 +1,26 @@
...
 from relay.network_graph.interests import calculate_interests
 
-InterestAccrued = namedtuple("InterestAccrued", ["value", "interest_rate", "timestamp"])
-
 DIRECTION_SENT = "sent"
 DIRECTION_RECEIVED = "received"
 
 
+@attr.s
+class InterestAccrued(object):
```
schmir-at-bb: There's no need to derive from `object` here. Please change it to not derive from `object`.

src/relay/blockchain/events_informations.py
```diff
@@ -0,0 +1,129 @@
...
+    elif most_recent_event_before_timestamp.direction == DIRECTION_RECEIVED:
+        if balance >= 0:
+            return most_recent_event_before_timestamp.interest_rate_received
+        else:
+            return most_recent_event_before_timestamp.interest_rate_given
+    else:
+        raise RuntimeError("Unexpected trustline update event")
+
+
+def sorted_events(events, reversed=False):
```
schmir-at-bb: Please rename the second argument to `reverse`, so it used the same name as `sorted`.

Issue #375, opened 2020-01-16
Add endpoint that returns all trustlines a user has in all currency networks indexed by the relay server


schmir-at-bb:
when done, add the functionality to the clientlib.

PR #385, opened 2020-01-21
List all trustlines

schmir-at-bb:
This is based on PR #384. Please only review the last two commits.

This implements #375.


cducrest:
I usually try to make `resources.py` call a single function of `relay.py` that does the whole computation in case we need to reuse the function in other places. It would also make `resources.py` unaware of the internals of `relay.py` and `currency_network_graphs`.

PR #388, opened 2020-01-24
Make the relay server sign transactions

schmir-at-bb:
To make the relay server sign transactions locally, you must configure
the account the be used via a snippet like the following in the TOML
configuration file:

,----
| [relay.account]
| keystore_path = "path/to/keystore.json"
| keystore_password_path = "path/to/passwd.txt"
`----

We use middleware that automatically signs transactions.

This requires the latest contract-deploy-tools.

This is an experimental feature. There might be race conditions with
the nonce handling, because previously we relied on parity to handle
the nonce.

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -13,7 +13,7 @@ certifi==2019.6.16
 cfgv==2.0.1
 chardet==3.0.4
 Click==7.0
-contract-deploy-tools==0.6.1
+contract-deploy-tools==0.7.1
 coverage==4.5.4
 cytoolz==0.10.0
 decorator==4.4.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,6 +30,6 @@ flake8-string-format
 pytest
 pytest-cov
 mypy
-contract-deploy-tools>=0.2.1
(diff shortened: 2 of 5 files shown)


schmir-at-bb:
> Can we maybe have tests to verify that if an account gets configured the transactions are actually sent from and are signed by that? I mean I'm pretty sure that the middlewares are added in the correct order and do work, but...

There was actually a bug hidden in there, that I didn't notice. So, I wrote some tests.

Besides as I already suspected, we had race conditions in there that I've already run into. I now do serialize the calls to ethSendTransaction.

> I would argue that the _install_w3_middle() function could be flatten for a cleaner flow to read, but that is personal a preference.

You have to write this middleware code with lot's of nested functions. I don't see a good way around that.

schmir-at-bb:
> To be honest I have problems to understand/follow the final big test case. But maybe I'm simply not deep enough in the topic. man_shrugging

It's just doing multiple transfers from the same address to the receiver with different combinations of parameters (both "from" and "nonce" supplied or not) and checks that the value has been transferred from the sender to the receiver.

PR #389, opened 2020-01-27
Remove pygraphviz

schmir-at-bb:
It doesn't look like it's still being.


schmir-at-bb:
> I think the removal should be complete. What was is used for in the past?

I don't know. Actually, looking at the history of the repository, it looks to me like it has never been used.

> I'm bit worried because the CI is doing nothing...

That's because I waited for it to succeed before opening a PR. github/circle will not show the circleci information in that case. I've told circleci to rebuild, and it's now there.

PR #390, opened 2020-01-27
Log the current greenlet

schmir-at-bb:
This provides useful information while debugging. Add relay.boot
module, similar to the bridge's boot module.


schmir-at-bb:
I've removed it because it's a new dependency. If we need it, we can add it later. I've used in the bridge in order to easily send signals to the bridge with killall.

Issue #92, opened 2018-05-24
make sure relay server errors will not go unnoticed by reporting them to sentry


schmir:
we need to make sure that errors will be noticed. Writing them to logfiles probably isn't enough. I did use sentry in the past to solve that.

Other alternatives would be Prometheus, Riemann, airbrake.

andfletcher:
@schmir can you add some ways to achieve this? Acceptance criteria or anything so that we can directly push this into refinement? Or make a suggestion to timebox this? Is this a global issue? Should we create a issue for error reporting in the project repo instead of the relay repo?

schmir:
Acceptance criteria would be: we get notified about certain errors without watching the log output of the relay server.

We can solve this by reporting errors to sentry.

At the moment I'd like to have this in the relay server, because the relay server will report errors to the client and we will not notice them.

schmir:
let's use sentry. In the first iteration we should hook sentry's error reporting into the flask web server and report errors for each HTTP request that returns a 5XX error code.

ToDo:

- [ ] add sentry as a dependency
- [ ] hook it up to flask web server
- [ ] make people register with our sentry account so they get notified via email

schmir-at-bb:
sentry is enabled for the staging server. Notifications per email do work.
I've invited Bernd as another owner of the sentry account. 

We still need to enable it on the production server.

PR #393, opened 2020-02-03
Report errors to sentry

schmir-at-bb:
This uses sentry's flask integration for reporting errors to
sentry. THis is a good first start, though we may need more reporting
in case an error doesn't happen during handling of a HTTP request or
we manually return an error code.

See https://github.com/trustlines-protocol/relay/issues/92


schmir-at-bb:
I'll just keep the original issue open.

PR #402, opened 2020-02-10
Provide more information on invalid meta-tx error message

cducrest:
closes: https://github.com/trustlines-protocol/relay/issues/401

### DIFF
--- a/src/relay/api/resources.py
+++ b/src/relay/api/resources.py
@@ -15,9 +15,13 @@ from relay.api import fields as custom_fields
 from relay.blockchain.currency_network_proxy import CurrencyNetworkProxy
 from relay.blockchain.delegate import (
     IdentityDeploymentFailedException,
+    InvalidChainId,
     InvalidDelegationFeesException,
     InvalidIdentityContractException,
     InvalidMetaTransactionException,
+    InvalidNonceHashPair,
+    InvalidSignature,
+    InvalidTimeLimit,
     UnknownIdentityFactoryException,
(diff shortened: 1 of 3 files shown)


src/relay/api/resources.py
```diff
@@ -440,12 +440,12 @@ def post(self, args):
                 400,
                 f"Invalid delegation fees: fees too low or not supported currency network of fees",
             )
-        except InvalidMetaTransactionException:
-            abort(400, "The meta-transaction is invalid")
-        except InvalidIdentityContractException:
+        except InvalidMetaTransactionException as exception:
+            abort(400, f"The meta-transaction is invalid: {exception}")
```
berndbohmeier: I think it is not good to print out an exception to a user, it might contain information we dont want to give out.

src/relay/blockchain/delegate.py
```diff
@@ -102,6 +85,35 @@ def validate_meta_transaction_fees(self, meta_transaction: MetaTransaction):
 
         raise InvalidDelegationFeesException()
 
+    def raise_on_invalid_meta_transaction(self, meta_transaction: MetaTransaction):
+        try:
+            if not self.delegate.validate_time_limit(meta_transaction):
+                raise InvalidMetaTransactionException(
+                    f"Invalid time limit for meta-tx: {meta_transaction.time_limit}"
```
berndbohmeier: I think this text that the user sees, should be in api.

PR #409, opened 2020-02-11
Add endpoint to ask whether identity was dpeloyed

cducrest:
Check if you like the endpoint name / url

closes https://github.com/trustlines-protocol/relay/issues/407


schmir-at-bb:
> If you really want this we should maybe at least use /is_identity_deployed/<address>

I would argue that this should take the address as query parameter then, since /is_identity_deployed/<address> doesn't refer to a resource anymore.

PR #426, opened 2020-02-19
Meta tx status endpoint

cducrest:
I don't know if I should use a schema to return the meta transaction status or simply use `MetaTransactionStatusField._serialize` without a schema.

requires: https://github.com/trustlines-protocol/contracts/pull/313

### DIFF
--- a/docs/RelayAPI.md
+++ b/docs/RelayAPI.md
@@ -1089,6 +1089,33 @@ curl --header "Content-Type: application/json" \
 
 ---
 
+### Status of meta transaction
+Get the status of a meta transaction for an identity.
+
+#### Request
+```
+GET /identities/:identityAddress/meta-transactions/:metaTransactionHash/status
+```
+
+#### Example Request
+```bash
+curl https://relay0.testnet.trustlines.network/api/v1/identities/0xF2E246BB76DF876Cef8b38ae84130F4F55De395b/meta-transactions/0x51a240271AB8AB9f9a21C82d9a85396b704E164d/status
+```
+
+#### Response
+| Attribute             | Type   | Description                                                       |
+|-----------------------|--------|-------------------------------------------------------------------|
+| status                | string | one of success, failure, pending, or not found                    |
+
+#### Example Response
+```json
+{
+  "status": "success"
+}
+```
+
+---
+
 ### Relay meta transaction
 Relays a meta transaction to the blockchain.
 #### Request
--- a/src/relay/api/app.py
+++ b/src/relay/api/app.py
@@ -41,6 +41,7 @@ from .resources import (
     Relay,
     RelayMetaTransaction,
     RequestEther,
+    Status,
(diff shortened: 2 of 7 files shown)


docs/RelayAPI.md
```diff
@@ -1089,6 +1089,33 @@ curl --header "Content-Type: application/json" \
...
+Get the status of a meta transaction for an identity.
+
+#### Request
+```
+GET /identities/:identityAddress/meta-transactions-status/:metaTransactionHash
+```
+
+#### Example Request
+```bash
+curl https://relay0.testnet.trustlines.network/api/v1/identities/0xF2E246BB76DF876Cef8b38ae84130F4F55De395b/meta-transactions/
```
berndbohmeier: the example seems to be wrong

docs/RelayAPI.md
```diff
@@ -1089,6 +1089,33 @@ curl --header "Content-Type: application/json" \
 
 ---
 
+### Status of meta transaction
+Get the status of a meta transaction for an identity.
+
+#### Request
+```
+GET /identities/:identityAddress/meta-transactions-status/:metaTransactionHash
```
berndbohmeier: should we not maybe have `meta-transaction` and than one field is status, so we can just add other fields later, like when it was included and so on.

src/relay/api/resources.py
```diff
@@ -492,6 +493,19 @@ def post(self, args):
...
 
 
+class MetaTransactionStatus(Resource):
+    def __init__(self, trustlines: TrustlinesRelay) -> None:
+        self.trustlines = trustlines
+
+    @dump_result_with_schema(MetaTransactionStatusSchema())
+    def get(self, identity_address: str, meta_transaction_hash):
+        return {
+            "MetaTransactionStatus": self.trustlines.get_meta_transaction_status(
```
berndbohmeier: together with the suggestion above i would just use `status` here

src/relay/api/app.py
```diff
@@ -138,6 +139,10 @@ def add_resource(resource, url):
         add_resource(DeployIdentity, "/identities")
 
     add_resource(IdentityInfos, "/identities/<address:identity_address>")
+    add_resource(
+        MetaTransactionStatus,
+        "/identities/<address:identity_address>/meta-transactions-status/<string:meta_transaction_hash>",
```
berndbohmeier: `meta-transactions` thats what you actually used in the clientlib pr

src/relay/api/schemas.py
```diff
@@ -79,6 +80,13 @@ class Meta:
     currencyNetworkOfFees = Address(required=True, attribute="currency_network_of_fees")
 
 
+class MetaTransactionStatusSchema(Schema):
+    class Meta:
+        strict = True
+
+    status = MetaTransactionStatusField(required="true")
```
berndbohmeier: "true"?

PR #427, opened 2020-02-20
Add function in node to get tx status

cducrest:
wip

### DIFF
--- a/src/relay/blockchain/node.py
+++ b/src/relay/blockchain/node.py
@@ -2,12 +2,18 @@ import logging
 import os
 from collections import namedtuple
 
+from tldeploy.identity import MetaTransactionStatus
+from web3.exceptions import TransactionNotFound
+
 from relay.concurrency_utils import synchronized
 
 TxInfos = namedtuple("TxInfos", "balance, nonce, gas_price")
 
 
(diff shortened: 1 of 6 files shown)


src/relay/blockchain/node.py
```diff
@@ -100,3 +103,20 @@ def send_ether(self, address):
 
     def get_block_timestamp(self, block_number):
         return self._web3.eth.getBlock(block_number).timestamp
+
+    def get_transaction_status(self, hash):
+        try:
+            # This will only find mined transactions and not pending ones
+            status = self._web3.eth.getTransactionReceipt(hash)["status"]
+            if status is not None:
```
berndbohmeier: this should not work, as it will not be set.
I would remove this check, as status should now always be set I think, no?

src/relay/blockchain/node.py
```diff
@@ -2,6 +2,9 @@
 import os
 from collections import namedtuple
 
+from tldeploy.identity import MetaTransactionStatus
```
berndbohmeier: I dont like this dependency. It does not make sense that normal transaction depend on MetaTransactionStatus. Just redefine it here as `TransactionStatus`

PR #429, opened 2020-02-20
Add valid default value for delegation fees

cducrest:
The endpoint for getting acceptable delegation fees would return an
empty list when no defaults were provided.
It should be better to return a fee with 0 value as an example of what
 will be accepted by the delegate.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -39,7 +39,7 @@ enable = true
 [delegate]
 enable = true
 enable_deploy_identity = true
-fees = []
+fees = [{'base_fee'= 0, 'currency_network'= '0x0000000000000000000000000000000000000000', 'gas_price'= 0}]
 
 # [account]
 # keystore_path = "path-to-keystor"
--- a/src/relay/config/schema.py
+++ b/src/relay/config/schema.py
@@ -6,6 +6,8 @@ from marshmallow import (
     pre_load,
 )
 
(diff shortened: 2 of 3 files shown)


src/relay/config/schema.py
```diff
@@ -28,7 +30,7 @@ class FeeSettingsSchema(Schema):
     base_fee = fields.Integer(missing=0)
     gas_price = fields.Integer(missing=0)
     fee_recipient = AddressField()
-    currency_network = AddressField(required=True)
+    currency_network = AddressField(missing=ZERO_ADDRESS)
```
schmir-at-bb: Please don't change that. If they miss to enter the address, they will be paid in the zero address currency network, i.e. not at all (which they may not have intended)

config.toml
```diff
@@ -39,7 +39,7 @@ enable = true
 [delegate]
 enable = true
 enable_deploy_identity = true
-fees = []
+fees = [{'base_fee'= 0, 'currency_network'= '0x0000000000000000000000000000000000000000', 'gas_price'= 0}]
```
berndbohmeier: This is not a good default as it makes no sense, please remove, I left it out on purpose

berndbohmeier:
I will open an issue, so we can solve it differenlty.

Issue #439, opened 2020-02-27
Extend documenation how to get contracts.json


weilbith:
From feedback of trustlines-network/project#812:
> Document how to get contract.json from npm or pypi instead of copying it out of a container.

PR #440, opened 2020-02-27
Add documentation how to get contracts.json file

weilbith:
Closes: #439

### DIFF
--- a/docker/trustlines/README.md
+++ b/docker/trustlines/README.md
@@ -74,14 +74,18 @@ reuse a keyfile from a validator node.
 
 ### Copy compiled contracts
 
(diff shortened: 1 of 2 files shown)


docker/trustlines/README.md
```diff
@@ -72,16 +74,6 @@ line.
 Of course you can also use an existing keyfile, but please do not
 reuse a keyfile from a validator node.
 
-### Copy compiled contracts
```
berndbohmeier: Please leave this I think, as this is the docker explanation this should be the default I guess. 
You can then mention here that there are other possibilities.

weilbith: :heavy_check_mark:

Issue #456, opened 2020-03-11
Update config of relay


cducrest:
- Default to not accept any meta-tx (fees list empty)
- When returning fees of 0, return `currencyNetworkOfFees = null`

PR #457, opened 2020-03-11
Feature/delegation fees config

cducrest:
Make config stricter for delegation fees: 
- if fees are 0 then no currency network should be set
- if fees are not 0 then a currency network should be set (to an address, we don't check what)
- if fess are in config, baseFee and gasPrice are required.
Make relay check for feeRecipient which breaks the e2e tests that will be fixed in the clientlib.

Closes https://github.com/trustlines-protocol/relay/issues/456

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -43,6 +43,8 @@ enable = true
 [delegate]
 enable = true
 enable_deploy_identity = true
+## The list of accepted fees for delegating a meta-tx
+## example for accepting 0 fees: [{'base_fee'=0, 'gas_price'=0}] other keys are 'currency_network' and 'fee_recipient'
 fees = []
 ## Set the maximum allowed gas usage per delegated meta transaction
 max_gas_limit = 1_000_000
--- a/config.toml
+++ b/config.toml
@@ -44,8 +44,10 @@ enable = true
 enable = true
 enable_deploy_identity = true
 ## The list of accepted fees for delegating a meta-tx
-## example for accepting 0 fees: [{'base_fee'=0, 'gas_price'=0}] other keys are 'currency_network' and 'fee_recipient'
-fees = []
+## Below is an example for accepting 0 fees. Other keys are 'currency_network' and 'fee_recipient'
+fees = [
+# {'base_fee'=0, 'gas_price'=0}
+]
 ## Set the maximum allowed gas usage per delegated meta transaction
 max_gas_limit = 1_000_000
 ## Supported gas price methods are rpc, fixed, bound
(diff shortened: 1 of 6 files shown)


config.toml
```diff
@@ -43,6 +43,8 @@ enable = true
 [delegate]
 enable = true
 enable_deploy_identity = true
+## The list of accepted fees for delegating a meta-tx
+## example for accepting 0 fees: [{'base_fee'=0, 'gas_price'=0}] other keys are 'currency_network' and 'fee_recipient'
```
berndbohmeier: Do we handle the fee recipient, do we check it?

config.toml
```diff
@@ -43,6 +43,8 @@ enable = true
 [delegate]
 enable = true
 enable_deploy_identity = true
+## The list of accepted fees for delegating a meta-tx
```
berndbohmeier: You can also write them as 
```
[[delegate.fees]]
base_fee = 0
gas_price = 0
```
in case you did not know

config.toml
berndbohmeier: It would be nice if we could have a test, that tests that our suggested example config is also working. 
I did that by removing single # in the example file, maybe you can adjust it, that it can also work int his case.

src/relay/blockchain/delegate.py
```diff
@@ -112,6 +111,7 @@ def validate_meta_transaction_fees(self, meta_transaction: MetaTransaction):
                 == meta_transaction.currency_network_of_fees
                 and fees_estimation.base_fee <= meta_transaction.base_fee
                 and fees_estimation.gas_price <= meta_transaction.gas_price
+                and fees_estimation.fee_recipient == meta_transaction.fee_recipient
```
berndbohmeier: How does that work, fee recipient does not have to be set in the config, no?
And fee recipient 0x00000 should be also fine, no?

tests/chain_integration/test_delegate.py
```diff
@@ -389,6 +389,33 @@ def test_meta_transaction_fees_invalid_network(
         )
 
 
+def test_meta_transaction_fee_recipient_invalid(
+    delegate_with_one_fees, signed_meta_transaction, owner_key
+):
+    """
+    Check that no exception is raised when validating a valid meta_transaction
```
berndbohmeier: ???

tests/unit/test_config.py
```diff
@@ -29,6 +29,24 @@ def uncommented_example_config_filepath(example_config_filepath, tmp_path):
     return str(d)
 
 
+@pytest.fixture()
+def correct_fees_config_file(tmp_path):
+    folder = tmp_path / "subfolder"
+    folder.mkdir()
+    file_path = folder / "wrong_config.toml"
```
berndbohmeier: why is it called wrong config?

src/relay/blockchain/delegate.py
cducrest: If it's not set in the config, it is using the node's address, isn't it?
https://github.com/trustlines-protocol/relay/blob/master/src/relay/relay.py#L338

We could also accept 0x000.

PR #458, opened 2020-03-11
[WIP] Use pip-tools to generate requirement files

berndbohmeier:
Use pip tools to generate requirement.txt and dev-requirements.txt.
This puts the version numbers in the requirement files and thus
makes the constraints.txt file unnecessary.
This allows us to more easily update our dependencies as this will now
be handled by pip-tools.


cducrest:
Why do we use `constraints.in` instead of setting the constraints in `setup.py`? Does that not mean that others installing the relay could not have the correct versions constrained?

Dockerfile
```diff
@@ -19,18 +19,15 @@ RUN apt-get update \
 RUN python3 -m venv /opt/relay
 
 WORKDIR /relay
-COPY ./constraints.txt /relay/constraints.txt
-RUN /opt/relay/bin/pip install --disable-pip-version-check -c constraints.txt pip wheel setuptools
+COPY ./dev-requirements.txt /relay/dev-requirements.txt
```
cducrest: Is there a reason to use `dev-requirements` instead of non-dev `requirements`?

.circleci/config.yml
```diff
@@ -38,24 +38,28 @@ jobs:
...
       - save_cache:
-          key: venv-{{ checksum "constraints.txt" }}-{{ checksum "requirements.txt" }}
+          key: venv-{{ checksum "dev-requirements.txt" }}-{{ checksum "requirements.txt" }}
           paths:
             - venv
       - run:
           name: Install package
           command: |
-            pip install -c constraints.txt -e .
+            pip install -e .
```
cducrest: Why do we not constraint?

tests/unit/network_graph/test_graph.py
```diff
@@ -213,7 +213,7 @@ def test_close_trustline_not_enough_capacity(
 
 
 def test_close_trustline_first_edge_insufficient_capacity(
-    complex_community_with_trustlines_and_fees
+    complex_community_with_trustlines_and_fees,
```
cducrest: :(

tests/unit/network_graph/test_graph.py
```diff
@@ -346,7 +349,10 @@ def test_capacity_path_single_hop_with_fees(complex_community_with_trustlines_an
     source = A
     destination = B
 
-    sendable, max_path = complex_community_with_trustlines_and_fees.find_maximum_capacity_path(
+    (
```
cducrest: What's up with black

.circleci/config.yml
berndbohmeier: What do you want to constrain? Everything is installed already. 
I can constrain it, but it should not change anything.

tests/unit/network_graph/test_graph.py
berndbohmeier: I guess they changes the rules. I updated black and that is what black then did to the code.

Dockerfile
berndbohmeier: This is for pip, wheel and setuptools, which we currently have defined as development requirements. I guess we should change that then, mmh

berndbohmeier:
The problem is that it is a sub-dependency. One of our dependencies is not correctly limiting its dependency. If I would put that limitation in our library, it would look like we directly depend on it and it would install it, even if it is not needed as sub dependency anymore.

PR #459, opened 2020-03-13
Fix/close path no path

cducrest:


cducrest:
Good point, I modified the tests to make it more obvious.

PR #472, opened 2020-04-02
Feature/transfer details from

cducrest:
Can get transfer details from txHash or blockHash + logIndex
Returns list of transfer details instead of error if there is multiple transfer in txHash

### DIFF
--- a/docs/RelayAPI.md
+++ b/docs/RelayAPI.md
@@ -808,21 +808,38 @@ The `accuredInterests` is a list with the following elements:
 ---
 
 ### Information for transfer in transaction
-Returns information about a trustline transfer applied by transaction with given hash.
+Returns information about all trustline transfer applied by transaction with given hash
+or identified by block hash and log index.
 #### Request
 ```
-GET /transfers/:txHash
+GET /transfers
 ```
+#### Data Parameters
+| Name            | Type   | Required | Description                                  |
+|-----------------|--------|----------|--------------------------------------------- |
+| transactionHash | string | No       | Hash of transaction responsible for transfer |
+| blockHash       | string | No       | Hash of block identifying transfer event     |
+| logIndex        | number | No       | Log index identifying transfer event         |
+
 #### Example Request
+To get transfer details via transaction hash:
+```bash
+curl --header "Content-Type: application/json" \
+  --request POST \
+  --data '{"transactionHash":"0x05c91f6506e78b1ca2413df9985ca7d37d2da5fc076c0b55c5d9eb9fdd7513a6"}' \
+https://relay0.testnet.trustlines.network/api/v1/transfers/
 ```
-curl https://relay0.testnet.trustlines.network/api/v1/transfers/0xC0B33D88C704455075a0724AA167a286da778DDE
+
+To get transfer details via transfer id:
+```bash
+curl --header "Content-Type: application/json" \
+  --request POST \
+  --data '{"blockHash":"0x05c91f6506e78b1ca2413df9985ca7d37d2da5fc076c0b55c5d9eb9fdd7513a6", "logIndex": 12}' \
+https://relay0.testnet.trustlines.network/api/v1/transfers/
 ```
-#### URL Parameters
-| Name         | Type                      | Required | Description                                       |
-|--------------|---------------------------|----------|---------------------------------------------------|
-| txHash       | string prefixed with "0x" | YES      | Hash of the tx responsible for the transfer       |
+
 #### Response
-The response is a an objects with the following elements:
+The response is a list of objects with the following elements:
 
 | Attribute        | Type                | JSON Type            | Description                                 |
 | ---------------- | ------------------- | -------------------- | ------------------------------------------- |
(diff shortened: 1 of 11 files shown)


src/relay/api/resources.py
```diff
@@ -476,19 +476,66 @@ class TransferInformation(Resource):
...
+                )
+            except TransferNotFoundException as e:
+                abort(
+                    404,
+                    f"No transfer found in transaction with transaction hash: {e.tx_hash}",
+                )
+        elif args["blockHash"]:
+            try:
+                return self.trustlines.get_transfer_information_from_event_id(
+                    args["blockHash"], args["logIndex"]
```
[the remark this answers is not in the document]
cducrest: This is not possible if we validate the input.

src/relay/api/resources.py
```diff
@@ -476,19 +476,66 @@ class TransferInformation(Resource):
...
+            except TransferNotFoundException as e:
+                abort(
+                    404,
+                    f"No transfer found in block {e.block_hash} with log index: {e.log_index}",
+                )
+            except IdentifiedNotTransferException as e:
+                abort(
+                    400,
+                    f"The event identified by block hash {e.block_hash} and log index {e.log_index} is not a Transfer",
+                )
```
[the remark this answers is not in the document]
cducrest: If the input is valid there is no else. Should I add an `else abort(500, "The relay server failed at validating the input.") ?

src/relay/blockchain/proxy.py
```diff
@@ -223,6 +223,42 @@ def get_transaction_events(
 
         return self._build_events(events)
 
+    def get_block_logs(self, block_hash):
+        return self._web3.eth.getLogs({"blockHash": block_hash})
+
+    def get_transaction_events_from_id(
```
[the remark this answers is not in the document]
cducrest: Ok

src/relay/api/resources.py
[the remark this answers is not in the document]
berndbohmeier: yes, but people make mistakes code changes, so please do `if args[blockhash] and args[logindex]`

src/relay/api/resources.py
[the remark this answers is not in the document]
berndbohmeier: raise something like `RuntimeError` or so, so that the error gets not eaten silently. Currently on a mistake we would just not return anything.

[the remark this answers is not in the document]
berndbohmeier: you dont need abort(500) this will happen automatically on unhandled exceptions

docs/RelayAPI.md
```diff
@@ -808,21 +808,38 @@ The `accuredInterests` is a list with the following elements:
...
+  --request POST \
+  --data '{"transactionHash":"0x05c91f6506e78b1ca2413df9985ca7d37d2da5fc076c0b55c5d9eb9fdd7513a6"}' \
+https://relay0.testnet.trustlines.network/api/v1/transfers/
 ```
-curl https://relay0.testnet.trustlines.network/api/v1/transfers/0xC0B33D88C704455075a0724AA167a286da778DDE
+
+To get transfer details via transfer id:
+```bash
+curl --header "Content-Type: application/json" \
+  --request POST \
```
berndbohmeier, line 836: below you define it as GET, and here as POST?

src/relay/api/schemas.py
```diff
@@ -280,3 +280,34 @@ class TransferInformationSchema(Schema):
...
+
+
+class TransferIdentifierSchema(Schema):
+    @validates_schema
+    def validate(self, data, partial, many):
+        transaction_hash = data["transactionHash"]
+        block_hash = data["blockHash"]
+        log_index = data["logIndex"]
+        if transaction_hash:
+            if block_hash or log_index:
```
berndbohmeier, line 292: any reason why you nest those?
I would just write it as
```
if transaction_hash and (blockhash or log_index)
```
because I think less nesting is preferable.

src/relay/blockchain/proxy.py
```diff
@@ -223,6 +223,44 @@ def get_transaction_events(
 
         return self._build_events(events)
 
+    def get_block_logs(self, block_hash):
+        return self._web3.eth.getLogs({"blockHash": block_hash})
+
+    def get_transaction_events_by_event_id(
+        self, block_hash, log_index, event_types: Tuple = None
+    ):
+        """Gets all all events from a transaction where event_id is the id of one of the events within the transaction
```
berndbohmeier, line 232: all all

src/relay/ethindex_db.py
```diff
@@ -488,3 +488,41 @@ def get_transaction_events(
         )
 
         return events
+
+    def get_transaction_events_by_event_id(
+        self, block_hash, log_index, event_types: Tuple = None
+    ):
+        """Gets all all events from a transaction
```
berndbohmeier, line 495: all all

src/relay/ethindex_db.py
```diff
@@ -488,3 +488,41 @@ def get_transaction_events(
...
+        event_types = self._get_standard_event_types(event_types)
+
+        query = EventsQuery(
+            """blockHash=%s
+               AND logIndex=%s
+            """,
+            (block_hash, log_index),
+        )
+
+        identified_event = self._run_events_query(query)[0]
```
berndbohmeier, line 508: is it handled if no event is found?

src/relay/api/schemas.py
cducrest: I think nesting is easier to understand, because it makes the following `elif` blocks more specific.

berndbohmeier: well me and zen of python do not agree, but its not that bad, so leave it.

cducrest: Too late

PR #473, opened 2020-04-02
Only get transfer and balance update events from tx for transfer details

cducrest:
This fixes a bug where the relay would not be able to get transfer
details for meta-transactions since they have DebtUpdate events in the
eth-index db and these have no corresponding event builders.

(I guess mypy helped me this time)

### DIFF
--- a/src/relay/blockchain/events_informations.py
+++ b/src/relay/blockchain/events_informations.py
@@ -74,7 +74,9 @@ class EventsInformationFetcher:
 
     def get_transfer_details(self, tx_hash):
 
-        all_events_of_tx = self.events_proxy.get_transaction_events(tx_hash)
+        all_events_of_tx = self.events_proxy.get_transaction_events(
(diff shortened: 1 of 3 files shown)


src/relay/blockchain/events_informations.py
```diff
@@ -74,7 +74,9 @@ def get_list_of_paid_interests_for_trustline_in_between_timestamps(
 
     def get_transfer_details(self, tx_hash):
 
-        all_events_of_tx = self.events_proxy.get_transaction_events(tx_hash)
+        all_events_of_tx = self.events_proxy.get_transaction_events(
+            tx_hash, standard_event_types=(TransferEventType, BalanceUpdateEventType)
```
berndbohmeier, line 78: why are they called `standard_event_types` and not just `event_types`?

In the other parts of the codes it was only named that way, because it was the default selected event types (mmh bettter name would have been maybe `default_event_types`) if you would not chose any, but here you chose.

Issue #419, opened 2020-02-17
Add event id to event rest  endpoints


berndbohmeier:
Add the event id as described in https://github.com/trustlines-network/project/issues/536 to the rest endpoints. for that you have to add it after the events are read from the database (or fetch it from the database, if already done, see: ().

andfletcher:
Reopend

berndbohmeier:
done

PR #475, opened 2020-04-03
Add log index and tx hash to returned events of rest endpoints

cducrest:
closes https://github.com/trustlines-protocol/relay/issues/419?


cducrest:
This is actually too annoying to have it separate from https://github.com/trustlines-protocol/relay/pull/472 as they will conflict and I need this to test getting transfer details from event id.

berndbohmeier:
Can you not just build 472 on top of this one?

PR #477, opened 2020-04-07
Add trustline events endpoint

### LINKED ISSUE #470: Add trustline events endpoint
Add an endpoint to receive all events related to a certain trustline

`/networks/:networkAddress/users/:userAddressA/trustlines/:userAddressB/events`

it should have the same options for filtering as the current ones, so `type` and `fromBlock`. 
also it should use the same default events as the other endpoints.

berndbohmeier:
Add an endpoint to query for the events of a trustlines. This includes
all relevent events, including the Balance update events, as they should
here be interesting.
closes #470

### DIFF
--- a/src/relay/api/resources.py
+++ b/src/relay/api/resources.py
@@ -345,6 +345,47 @@ class UserEventsNetwork(Resource):
             abort(504, TIMEOUT_MESSAGE)
 
 
+class TrustlineEvents(Resource):
+    def __init__(self, trustlines: TrustlinesRelay) -> None:
+        self.trustlines = trustlines
+
+    args = {
+        "fromBlock": fields.Int(required=False, missing=0),
+        "type": fields.Str(
+            required=False,
+            validate=validate.OneOf(CurrencyNetworkProxy.trustline_event_types),
+            missing=None,
+        ),
+    }
+
+    @use_args(args)
+    @dump_result_with_schema(UserCurrencyNetworkEventSchema(many=True))
+    def get(
+        self, args, network_address: str, user_address: str, counter_party_address: str
+    ):
+        abort_if_unknown_network(self.trustlines, network_address)
+        from_block = args["fromBlock"]
+        type = args["type"]
+        try:
+            return self.trustlines.get_trustline_events(
+                network_address,
+                user_address,
+                counter_party_address,
+                type=type,
+                from_block=from_block,
+            )
+        except TimeoutException:
+            logger.warning(
+                "Trustline events: event_name=%s user_address=%s "
+                "counter_party_address=%s, from_block=%s. could not get events in time",
+                type,
+                user_address,
+                counter_party_address,
+                from_block,
+            )
+            abort(504, TIMEOUT_MESSAGE)
+
+
 class UserEvents(Resource):
     def __init__(self, trustlines: TrustlinesRelay) -> None:
         self.trustlines = trustlines
--- a/src/relay/ethindex_db.py
+++ b/src/relay/ethindex_db.py
@@ -2,7 +2,7 @@
 
 import collections
 import logging
-from typing import Any, Dict, List, Tuple
+from typing import Any, Dict, Iterable, List, Optional
 
 import psycopg2
 import psycopg2.extras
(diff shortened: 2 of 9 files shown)


src/relay/api/resources.py
```diff
@@ -344,6 +344,45 @@ def get(self, args, network_address: str, user_address: str):
...
+            return self.trustlines.get_trustline_events(
+                network_address,
+                user_address,
+                counter_party_address,
+                type=type,
+                from_block=from_block,
+            )
+        except TimeoutException:
+            logger.warning(
+                "Trustline events: event_name=%s user_address=%s from_block=%s. could not get events in time",
```
cducrest, line 378: You could log counter_patry_address (also maybe we should uniformly use counterparty in the code)

src/relay/ethindex_db.py
```diff
@@ -255,49 +257,34 @@ def get_trustline_events(
...
+
+        all_from_to_types = set()
+        for from_, to in self.from_to_types.values():
+            all_from_to_types.add((from_, to))
+            all_from_to_types.add((to, from_))
+
+        from_to_where = " OR ".join(
+            f"(args->>'{from_}'=%s AND args->>'{to}'=%s)"
+            for from_, to in all_from_to_types
+        )
```
cducrest, line 274: I would use `a` and `b` or `user` and `counterparty` since they are not `from` and `to` anymore but could be `to` and `from`.

Also I don't like the name `from_to_where`, but maybe it's an SQL thing. Would prefer something like `from_to_request_filter` or `from_to_filter`

src/relay/relay.py
```diff
@@ -175,10 +177,22 @@ def get_event_selector_for_currency_network(self, network_address=None):
...
+        """
+        if self.use_eth_index:
+            return ethindex_db.EthindexDB(
+                ethindex_db.connect(""),
+                standard_event_types=currency_network_events.standard_event_types,
+                event_builders=currency_network_events.event_builders,
+                from_to_types=currency_network_events.from_to_types,
+            )
+        else:
+            return Proxy(self._web3, abi=self.contracts["CurrencyNetwork"]["abi"])
```
cducrest, line 194: I don't think it is clear that you ask for a general proxy and get a proxy which only knows the abi of currency networks.

src/relay/relay.py
```diff
@@ -522,13 +534,27 @@ def get_trustline_events(
...
+        if self.use_eth_index:
+            proxy = ethindex_db.EthindexDB(
+                ethindex_db.connect(""),
+                address=network_address,
+                standard_event_types=currency_network_events.trustline_event_types,
+                event_builders=currency_network_events.event_builders,
+                from_to_types=currency_network_events.from_to_types,
+            )
+        else:
+            proxy = self.currency_network_proxies[network_address]
```
cducrest, line 551: This is a duplication of `get_event_selector_for_currency_network` except for `standard_event_types`, I don't fully get the purpose of `standard_event_types` right now but could this be avoided?

src/relay/ethindex_db.py
berndbohmeier: WHERE is the sql keyword for filter, so thats why I use where

src/relay/relay.py
berndbohmeier: Idk, I tried to clean up some of the mess that we sometimes dont return a CurrencyNetworkProxy in the `get_event_selector_for_currency_network`
I'll try to give it a better name.

src/relay/relay.py
berndbohmeier: This will be cleaned up anyways when we remove the proxies, so I think it is fine for now.

src/relay/api/resources.py
berndbohmeier: will do

berndbohmeier: And for the counter_party this will be tough, as we use counter_party or counterParty everywhere.

berndbohmeier: If you care, please open an issue

cducrest: I don't care, but I thought we used both counterparty and counter_party and wanted to use only one of them.

src/relay/relay.py
cducrest: I like the new name better.

PR #482, opened 2020-04-09
Squash together two queries to get transaction events

cducrest:


src/relay/ethindex_db.py
```diff
@@ -481,30 +481,16 @@ def get_transaction_events_by_event_id(
...
 
         event_types = self._get_standard_event_types(event_types)
 
         query = EventsQuery(
-            """blockHash=%s
-               AND logIndex=%s
+            f"""blockHash=%s
+            AND eventName in %s
+            AND transactionHash IN
+                (SELECT transactionHash "transactionHash" FROM events WHERE blockHash=%s AND logIndex=%s)
```
berndbohmeier, line 491: add `LIMIT 1` or so to express that we only need one result

PR #497, opened 2020-04-17
Refactor ethindex

cducrest:
Based on: https://github.com/trustlines-protocol/relay/pull/495

I found it hard to make refactoring that was straight improvement. I hope I did the most that is not so opinionated.

I thought about removing `address` from EthindexDB because it is confusing as to which method will use a default address for filtering, but I did not think it was an obvious improvement since then all methods using the default address will need to have it as a non-optional parameter.

I had the choice to remove all `get_ethindex_db_for_...` functions and have specific classes for each contracts that inherit `EthindexDB` with static `standard_event_types`, `from_to_types` and `event_builders`,  as it is the main difference in between each of these functions. I thought that was not obviously better.

I think `get_all_contract_events` is dirty as it does too many different things depending on its arguments, and it could lead to mistake if used without `contract_address`, but it is challenging to fix it since it is used in many places. https://github.com/trustlines-protocol/relay/pull/497/files#diff-6caecb6c151ad021cb4ee3f86396524dL340

### DIFF
--- a/src/relay/api/resources.py
+++ b/src/relay/api/resources.py
@@ -341,7 +341,7 @@ class UserEventsNetwork(Resource):
             )
         except TimeoutException:
             logger.warning(
-                "User network events: event_name=%s user_address=%s from_block=%s. could not get events in time",
+                "User network events: event_type=%s user_address=%s from_block=%s. could not get events in time",
                 type,
                 user_address,
                 from_block,
@@ -380,7 +380,7 @@ class TrustlineEvents(Resource):
             )
         except TimeoutException:
             logger.warning(
-                "Trustline events: event_name=%s user_address=%s "
+                "Trustline events: event_type=%s user_address=%s "
                 "counter_party_address=%s, from_block=%s. could not get events in time",
                 type,
                 user_address,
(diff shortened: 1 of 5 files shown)


src/relay/api/resources.py
```diff
@@ -341,7 +341,7 @@ def get(self, args, network_address: str, user_address: str):
             )
         except TimeoutException:
```
berndbohmeier, line 342: We can probably remove the time out stuff, as this should be related to the old contract code

cducrest: I think it is true for all of them except this one: https://github.com/trustlines-protocol/relay/pull/497/files#diff-a005b72d629c8034ee162be425c13202R232
It spawns greenlet for each exchange addresses and make a query to an ethindex_db instance for each of them (one db instance by addresses), so it can time out.

src/relay/relay.py
```diff
@@ -582,58 +582,6 @@ def get_user_events(
...
-                queries.append(
-                    functools.partial(
-                        ethindex_db.get_all_unw_eth_events,
-                        user_address=user_address,
-                        from_block=from_block,
-                    )
-                )
-        return queries
-
     def _get_exchange_event_queries(
```
berndbohmeier, line 585: maybe add a todo to remove this?

Issue #488, opened 2020-04-14
Create endpoint to query for delegation fees of meta-transaction


berndbohmeier:
Create an endpoint to ask for delegation fees. It should take transactionHash as parameter, but metaTransactionHash and identityAddress should be easy to add later.

PR #498, opened 2020-04-23
Add endpoint to get paid delegation fees via tx hash

cducrest:
Closes: https://github.com/trustlines-protocol/relay/issues/488

I attempted to solve the problem while making only one query to the database and not receiving too much information. This has the drawback that if there is more than one `DebtUpdate` event in the identifies transaction it does not work (we could probably make it work with better postgresql knowledge). But I believe the reason why you would have more than one such event is if the delegate is evil, and he can attack you in other way if he is evil anyway.

Also includes the removal of the remaining `timeout` in event queries and the removal of `event_query_timeout` from the config since it is not used anymore.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -9,7 +9,6 @@ gas_price = 0
 [trustline_index]
 enable = true
 full_sync_interval = 300
-event_query_timeout = 20
 
 [tx_relay]
 enable = true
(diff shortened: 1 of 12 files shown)


cducrest:
We deicided to implement the feature using the node to query for event to use the identity's `feePayment` event instead of the currency network's `debtUpdate`.

PR #500, opened 2020-04-28
Feature/get paid fees from node

cducrest:
closes https://github.com/trustlines-protocol/relay/issues/488
Breaks e2e tests that will be fixed by https://github.com/trustlines-protocol/end2end/pull/26 and an upcoming clientlib PR

### DIFF
--- a/src/relay/blockchain/identity_events.py
+++ b/src/relay/blockchain/identity_events.py
@@ -1,0 +1,26 @@
+from .events import BlockchainEvent
+
+FeePaymentEventType = "FeePayment"
+
+
+class FeePaymentEvent(BlockchainEvent):
+    @property
+    def value(self):
+        return self._web3_event.get("args").get("value")
+
+    @property
+    def from_(self):
+        return self._web3_event.get("address")
+
+    @property
+    def to(self):
+        return self._web3_event.get("args").get("recipient")
+
+    @property
+    def currency_network(self):
+        return self._web3_event.get("args").get("currencyNetwork")
+
+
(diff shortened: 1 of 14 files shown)


src/relay/blockchain/identity_events.py
```diff
@@ -0,0 +1,26 @@
...
+FeePaymentEventType = "FeePayment"
+
+
+class FeePaymentEvent(BlockchainEvent):
+    @property
+    def value(self):
+        return self._web3_event.get("args").get("value")
+
+    @property
+    def from_(self):
```
weilbith, line 12: Not part of the review, but could anybody point me to a document or explain
me himself what the convention about the `_` usage is? I only know these:
- in Python a leading underscore signals the intended module scope
- in Solidity a leading underscore signals a local function parameter
- in Solidity a following underscore is used to avoid name collisions

So how do we use it here?

cducrest: You can't name a parameter `from` in python because it means a `from` loop.
If you would want to do that, you'd write `from = 'whatever'` and python would think you start a `from` loop and then not expect the following `=` sign.

cducrest:
Regarding the renaming of the log parameter name, we do not have an `arg` of the log for the debtor, so I had to use `address`, which I thought would be good to rename. So I thought I would pick the easiest names to understand `from_` and then renamed the creditor to `to`.

This is usually abstracted in events by the use of `from_to_types` which is not possible here since there is no `from` type.

PR #501, opened 2020-04-28
Add extra data to transfer details

cducrest:


weilbith (approved):
Is this data already available by the indexer?
Else should be fine. :+1:

PR #504, opened 2020-05-11
Small event related refactors

berndbohmeier:


cducrest (approved):

docker/trustlines/config.toml
```diff
@@ -7,7 +7,6 @@ gas_price = 0
 
 [trustline_index]
 full_sync_interval = 300
-event_query_timeout = 20
```
cducrest, line 10: Shouldn't you remove that from the config schema? Otherwise people might still change this value in their config and expect some change (idk what).

berndbohmeier: did you not remove that already? (https://github.com/trustlines-protocol/relay/commit/05ef9a30a58624e7982d5ecaf95ff8f6a97b5bd0#diff-8f9b1c5d647994a35c2fa47dbe91b857) 
or what do you mean?

Issue #75, opened 2018-04-18
Unlist currency networks that are removed from the addresses file


berndbohmeier:
Currently the relay will add new currency networks when they are added in the addresses.json file but they are not removed once they are removed from the file. 
Biggest issue: We need to also stop listening for events.

PR #510, opened 2020-06-22
No longer listen on new addresses for networks / exchanges ...

cducrest:
Deprecate the relay config for the update indexed networks interval
that is relay.updateNetworksInterval and
relay.update_indexed_networks_interval

closes https://github.com/trustlines-protocol/relay/issues/75

### DIFF
--- a/src/relay/config/config.py
+++ b/src/relay/config/config.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Dict, List, MutableMapping, Optional, Union
+from typing import Dict, Iterable, List, MutableMapping, Union
 
 import toml
 from marshmallow import ValidationError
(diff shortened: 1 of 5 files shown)


src/relay/config/config.py
```diff
@@ -105,6 +104,8 @@ def convert_delegation_fees(old_fees):
     not_supported_anymore_config: Dict[str, Optional[str]] = {
         "trustline_index.event_query_timeout": None,
         "relay.eventQueryTimeout": None,
+        "relay.updateNetworksInterval": None,
```
berndbohmeier, line 107: Lets deprecate it first

src/relay/relay.py
```diff
@@ -733,21 +729,6 @@ def _start_listen_network(self, address):
             self._process_network_freeze, start_log_filter=False
         )
```
berndbohmeier, line 731: I think you have to at least call it once.

PR #526, opened 2020-07-23
Feature/data messages

### LINKED ISSUE #490: Upgrade firebase admin to 4.x
The upgrade needs some code changes, see https://github.com/firebase/firebase-admin-python/releases

### LINKED ISSUE #515: Include event id in push notification payload
As the app guys want to change their push notification processing we have to change what we sent. 
They want to process the messages themselves, so we will only sent data notifications. 

- [x] Remove notification part and only sent data
- [x] Include all the data to identify an event (blockHash, logIndex and transactionHash)
- [x] Include data to check when to expect the event (blockNumber)
- [x] Include data to help them open the correct screen (eventType)
- [x] set android priority high (was claimed that this is needed, maybe doublecheck)

I think it would good to follow the data structure of event return by the event endpoints for consistency. 

Something like

```
{
  "message": {
    "data": {
      "blockHash": "012356674774747474747474",
      "eventType": "TrustlineUpdateRequest",
    },
    "android": {
      "priority": "HIGH"
    }
  }
}
```
or maybe have it under `data.event`

compojoom:
fixes #515 and fixes #490


src/relay/pushservice/pushservice.py
```diff
@@ -116,41 +120,45 @@ def check_client_token(self, client_token: str) -> bool:
...
-            )
+            data = {"message": event.message, "eventType": event.type}
+
+    return data
+
 
-    return notification
+def _get_data_prop_dict(event):
+    return {
+        "blockHash": str(event.block_hash.hex()),
```
cducrest, line 159: You may want to prepend `0x`, I don't think it will be there otherwise.

src/relay/relay.py
```diff
@@ -733,7 +735,10 @@ def _start_listen_network(self, address):
     def _start_push_service(self):
         logger.info("Start pushnotification service")
         path = self.config["push_notification"]["firebase_credentials_path"]
-        self._firebase_raw_push_service = FirebaseRawPushService(path)
+        cred = credentials.Certificate(path)
+        self._firebase_raw_push_service = FirebaseRawPushService(
+            firebase_admin.initialize_app(cred)
+        )
```
cducrest, line 741: Why do you think it is better to initialize the firebase app in `relay` than in `FirebaseRawPushService`?

tests/unit/pushnotifications/test_pushservice.py
```diff
@@ -0,0 +1,167 @@
...
+def _instrument_messaging_service(app, status=200, payload=_DEFAULT_RESPONSE):
+    fcm_service = messaging._get_messaging_service(app)
+    recorder = []
+    fcm_service._client.session.mount(
+        "https://fcm.googleapis.com", MockAdapter(payload, status, recorder)
+    )
+    return fcm_service, recorder
+
+
+class TestSend:
```
cducrest, line 50: You don't need this class, is it just a way to group tests that tests similar behaviour together? We could do that, but then you need to do that for the other tests of this module.

src/relay/relay.py
compojoom: Yes, I was unsure about this, but couldn't find a better way than dependency injection. 

The problem is that in order to test the firebaseRawPushService I had to provide a path to the firebase config file.  In unit tests one needs to mock the firebase credentials. So my first implementation was - create the credentials object and pass it to the rawpushservice. Tehn however to further test the sending one needs to set a dummy project for the firebase_app and since the app is inside of the firebase push service I couldn't easily do that. So I decided to create the firebase app outside of the raw push service and inject it. This way I can test the implementation with unit tests. 

Better ideas?

src/relay/pushservice/pushservice.py
compojoom: the 0x should be there - otherwise this here https://github.com/trustlines-protocol/relay/pull/526/files#diff-d98cf108481d4c897e05c4cf013b9279R131 shouldn't have passed?

src/relay/pushservice/pushservice.py
```diff
@@ -54,18 +53,23 @@ def dedup_event_id(client_token, event: Event):
...
+            app: The initialized firebase_admin App
         """
-        cred = credentials.Certificate(path_to_keyfile)
-        self._app = firebase_admin.initialize_app(cred)
+
+        if not isinstance(app, firebase_admin.App):
+            raise Exception("Firebase admin app not initialized")
+
+        self._app = app
+        print(self._app)
```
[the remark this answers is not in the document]
compojoom: ooops.

src/relay/relay.py
compojoom: I created a helper function that can initialize the firebase app from the provided path. This way we are moving the credentials dependency outside of the relay.py.

src/relay/pushservice/pushservice.py
```diff
@@ -51,21 +50,33 @@ def dedup_event_id(client_token, event: Event):
...
         """
         Initializes the push service
         Args:
-            path_to_keyfile: Path to json keyfile with firebase credentials
+            app: The initialized firebase_admin App
         """
-        cred = credentials.Certificate(path_to_keyfile)
-        self._app = firebase_admin.initialize_app(cred)
+
+        if not isinstance(app, firebase_admin.App):
```
berndbohmeier, line 71: I would not do such a strict check, as it destroys the flexibility of python to just pass in something wich provides the right methods.

tests/unit/pushnotifications/test_pushservice.py
```diff
@@ -0,0 +1,167 @@
...
+    fcm_service = messaging._get_messaging_service(app)
+    recorder = []
+    fcm_service._client.session.mount(
+        "https://fcm.googleapis.com", MockAdapter(payload, status, recorder)
+    )
+    return fcm_service, recorder
+
+
+def test_send_on_blockchain_event(web3_event_transfer):
+    raw_service = FirebaseRawPushService(app=admin_app)
```
berndbohmeier, line 51: I would have put them (raw_service and recorder) also in a fixture, but it is a minor detail.

tests/unit/conftest.py
```diff
@@ -0,0 +1,68 @@
...
+                "_isFrozen": True,
+            },
+            "event": TrustlineUpdateEventType,
+        }
+    )
+    return web3_event
+
+
+@pytest.fixture()
+def web3_event_trustline_request(web3_event):
```
berndbohmeier, line 39: It could lead to issue if you want to use two events in the same test, because we update the event here. just as a note.

compojoom: .copy() it is!

compojoom:
Tested locally and messages are properly delivered to iOS devices.

PR #534, opened 2020-09-02
Add `PaymentRequestDecline` msg type to push service

dakingha69:
The message of type `PaymentRequestDecline` was not whitelisted. Therefore no push notification was triggered which is fixed with this.

I tested it locally with the end2end setup and apps connected to the local relay.

### DIFF
--- a/src/relay/pushservice/pushservice.py
+++ b/src/relay/pushservice/pushservice.py
@@ -153,7 +153,7 @@ def _build_data_prop(event: Event):
         if event.direction == "received":
             data = _get_data_prop_dict(event)
     elif isinstance(event, MessageEvent):
-        if event.type == "PaymentRequest":
+        if event.type in ["PaymentRequest", "PaymentRequestDecline"]:
             data = {"message": event.message, "eventType": event.type}
 
     return data
--- a/tests/unit/pushnotifications/test_pushservice.py
+++ b/tests/unit/pushnotifications/test_pushservice.py
@@ -28,8 +28,19 @@ requestMessagePayload = (
     '"user":"0xE85bd548b2C961A2385628dcbBcc9A2E480dD925"} '
 )
 
-message_event = MessageEvent(
+requestDeclineMessagePayload = (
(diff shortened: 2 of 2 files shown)


tests/unit/pushnotifications/test_pushservice.py
```diff
@@ -84,9 +95,11 @@ def test_send_on_blockchain_event(raw_push_service, recorder, web3_event_transfe
 
 
 def test_send_on_non_blockchain_event(raw_push_service, recorder):
-    raw_push_service.send_event(client_token="token", event=message_event)
+    raw_push_service.send_event(
+        client_token="token", event=payment_request_message_event
+    )
 
-    assert_body_has_correct_payload(recorder, message_event)
+    assert_body_has_correct_payload(recorder, payment_request_message_event)
```
cducrest, line 102: You could parametrize this test to also run with the decline message event

src/relay/pushservice/pushservice.py
```diff
@@ -153,7 +153,7 @@ def _build_data_prop(event: Event):
         if event.direction == "received":
             data = _get_data_prop_dict(event)
     elif isinstance(event, MessageEvent):
-        if event.type == "PaymentRequest":
+        if event.type in ["PaymentRequest", "PaymentRequestDecline"]:
```
cducrest, line 156: Are we sure that's the two only event type we want to build data for?
Why do we not build data for every event type no matter what they are?

dakingha69: Yes currently they should be the only ones. The remaining message types are for example `PaymentMessage` and `RemoteNameMessage`, which are currently not handled as push notifications.

tests/unit/pushnotifications/test_pushservice.py
dakingha69: ok will do

dakingha69: done

cducrest (approved):

PR #535, opened 2020-09-03
Update the exceptions thrown by firebase in pushnotifications

cducrest:
They changed following firebase admin sdk  version 3
https://firebase.google.com/docs/admin/migrate-python-v3


dakingha69 (approved):
LGTM

Issue #502, opened 2020-05-04
Wrong endpoint / docs for user events


cducrest:
We document an endpoint to get events of all user in all currency networks: https://github.com/trustlines-protocol/relay/blob/master/docs/RelayAPI.md#events-of-user-in-all-currency-networks
But it seems this endpoint actually returns all events for the user whether they are currency network events or not.
https://github.com/trustlines-protocol/relay/blob/master/src/relay/api/resources.py#L373

More importantly I would find it hard to implement a way to get all currency network events for a user from the database. Since there are `Transfer` in currency networks but also in tokens, filtering by event types does not work and we need to filter by address of known currency networks.

Could there be a way to change the way we index events to add the contract type to the event?

PR #542, opened 2020-09-23
Enhance endpoint for user events to add filtering by contract type

cducrest:
It was documented to provide events for currency networks, but was
actually returning events for any contracts (exchange, tokens ...)
Closes https://github.com/trustlines-protocol/relay/issues/502


dakingha69 (approved):
LGTM 👍 Only small typos

PR #545, opened 2020-10-07
Feature/tests with database

cducrest:
Setups of tests with running a database with docker, and the py-eth-index.
Needs a new version of contract-deploy-tools that allows to expose the internal test chain used in our tests via rpc.


compojoom (approved):
LGTM. Tested it on a mac and it works fine as well. 👍

PR #547, opened 2020-10-19
Feature/meidation fees in time window

cducrest:
Based on https://github.com/trustlines-protocol/relay/pull/546
closes https://github.com/trustlines-protocol/relay/issues/348

PR #548, opened 2020-10-21
Feature/delegate debt

cducrest:
related to: https://github.com/trustlines-network/project/issues/752
based on https://github.com/trustlines-protocol/relay/pull/547

Endpoint to return list of debts of user.

There is no test for the api, since we never had tests for the api in the relay and mostly rely on the clientlib e2e test. However this user story does not (so far) require a clientlib function to call the endpoint. Thus there will be no test for the endpoint. I tested it manually.

To test it: 
- build the relay docker: `docker build -t trustlines/relay .`
- start the e2e tests backend with fees: `../end2end/run-e2e.sh -f -b`
- start some clientlib end2end tests to get some debts toward delegate `nyc mocha --timeout 30000 --require isomorphic-fetch -r ts-node/register tests/e2e/Trustline.test.ts`
- query for debt list: `curl localhost:5000/api/v1/users/0x00a329c0648769A73afAc7F9381E08FB43dBEA72/debts`

Missing to close user story is find whether there is a path from creditor to debtor.

### DIFF
--- a/docs/RelayAPI.md
+++ b/docs/RelayAPI.md
@@ -43,6 +43,7 @@ https://relay0.testnet.trustlines.network/api/v1
 - [Transaction infos for user](#transaction-infos-for-user)
 - [Balance of user](#balance-of-user)
 - [All trustlines](#all-trustlines-of-user-in-all-currency-networks)
+- [All debts of user](#all-debts-of-user)
 ### Other
 - [Latest block number](#latest-block-number)
 - [Relay transaction](#relay-transaction)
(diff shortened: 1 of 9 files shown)


docs/RelayAPI.md
```diff
@@ -1006,6 +1007,66 @@ The `paymentPath` is an object with the following attributes:
 
 ---
 
+### Earned mediation fees of user
+Returns a list of all earned mediation fees for user in a network.
+#### Request
```
compojoom, line 1012: Before explaining what the endpoint does can you add a short description of what "mediation fees" are?

In the app we have "imbalance fees", "delegation fees" - now we talk about "mediation fees". I think I know what mediation fee is supposed to be, but an outsider will have difficulties with the domain language.

compojoom (approved):

PR #549, opened 2020-10-23
Feature/delegate debt path

cducrest:
Based on https://github.com/trustlines-protocol/relay/pull/548
closes https://github.com/trustlines-network/project/issues/752
Add the information about the maximum claimable debt and path

### DIFF
--- a/src/relay/relay.py
+++ b/src/relay/relay.py
@@ -304,7 +304,11 @@ class TrustlinesRelay:
     def get_debt_list_of_user(self, user_address):
 
         event_selector = self.get_ethindex_db_for_currency_network()
-        return EventsInformationFetcher(event_selector).get_debts_lists(user_address)
+        return EventsInformationFetcher(
+            event_selector
+        ).get_debt_lists_in_all_networks_with_path(
+            user_address, currency_network_graphs=self.currency_network_graphs
+        )
 
     def deploy_identity(self, factory_address, implementation_address, signature):
         return self.delegate.deploy_identity(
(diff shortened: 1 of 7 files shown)


src/relay/relay.py
```diff
@@ -304,7 +305,48 @@ def get_earned_mediation_fees(
...
+                    capacity_path.capacity, abs(debt)
+                )
+                if capacity_path.path != []:
+                    enriched_debt["claim_path"] = capacity_path.path
+
+                enriched_debts_list.append(enriched_debt)
+
+            enriched_debts_lists_in_all_currency_networks.append(
+                {"currency_network": currency_network, "debts": enriched_debts_list}
+            )
```
compojoom, line 347: Move this block to a separate function. addAdditionalDebtInformation or I don't know what.

Then your code would read like poetry :)
...
get_debt_list_of_user
first fetch the user's debts list
for each debt add debtor, value and path

the end...

And the person reading on the other hand wouldn't have to go into the internal details of the for loop, but will still know what you do there :)

cducrest: I didn't want to do that because the structure of `debts_list_in_all_currency_networks` is kind of implicit, it depends on `EventsInformationFetcher.get_debts_lists`. And the structure of the output of this function depends on the schema `DebtsListInCurrencyNetwork` (or the other way around if you will). So I didn't want to split it and separate even more the responsibilities while it is actually a bit entangled.

cducrest: I'll rework it.

cducrest:
I guess backwards compatibility is broken because the previous e2e tests were using interests rates of 300% while the current contract version limits to 20%.

We shall now run the backwards compatibility checks to newest clientlib e2e test version.

compojoom (approved):

PR #552, opened 2020-10-27
Remove empty interests from endpoint of accured interest for user

cducrest:
We were returning an empty list object when there was no accrued
interests with the friend of a user, return none instead


compojoom (approved):
nice catch!

PR #557, opened 2020-11-18
Feature/graph sync from ethindex

cducrest:


src/relay/ethindex_db/sync_updates.py
```diff
@@ -0,0 +1,97 @@
...
+
+def get_graph_updates_feed(
+    conn,
+) -> List[Union[TrustlineUpdateFeedUpdate, BalanceUpdateFeedUpdate]]:
+    """Get a list of updates to be applied on the trustlines graphs to make them up to date with the chain"""
+
+    query_string = """
+        WITH deleted as (DELETE FROM graphfeed * RETURNING *)
+        SELECT * FROM deleted ORDER BY id ASC;
+    """
```
compojoom, line 71: As we discussed this - I don't think that we can do this in production. It's a huge anti-pattern to do something like this and we should find a better way.

compojoom (approved):
I think that it is good now. 
This approach solves our current issue, but also comes with its own drawbacks.

Main problems - how do you sync the relay was offline for a week, but the py-eth-index was not and we got a lot of trustline updates in the meantime. 

We'll monitor this and we'll see if it proves to be a problem.

One possible solution if the eth-indexer is only used with one relay would be to prune the relay graphfeed table when starting the relay.

PR #562, opened 2020-12-16
Update docs to point to new dev. docs website

compojoom:
The source of truth for all developer docs is the dev docs website: https://dev.trustlines.network

I’ve updated the README’s to point to it and removed the docs that were stored within the repository.

The repo that holds the docs is here:
https://github.com/trustlines-protocol/tl-dev-docs/tree/master/relay


cducrest (commented):
Why do we keep the contributing.md here and not on the dev docs site, while we keep the pre-commit instructions, and how to release there for example?

cducrest (approved):

compojoom:
There is a contributing page on the dev docs page. I'm thinking however of actually removing it and just linking to the Contributing.md in the repo.

I thought that it is good to have the contributing section in the readme in order to make it more visible that we accept PRs from external contributors. That's all.

PR #566, opened 2021-01-28
The hyperlinks in the readme were wrongly formatted

cducrest:
That would prevent us from uploading the new relay version to pypi


compojoom (approved):
omg

PR #571, opened 2021-02-26
Properly read the last graph sync id from file when there is no comment

cducrest:


src/relay/ethindex_db/sync_updates.py
```diff
@@ -150,6 +150,16 @@ def get_latest_graph_sync_id():
     with open(SYNC_FILE_PATH, "r") as f:
         contents = f.read()
 
-    # first line is a comment
-    # second line which we return is the last synced id
-    return contents.split("\n")[1]
+    lines = contents.split("\n")[1]
```
compojoom, line 153: You sure you want to do this way?
Wouldn't it be easier to just give up on the comment, add the id as first thing in the file. Then when the server crashes with cannot write file whatever -> the user can just create and doesn't have to deal with anything else.

compojoom: Maybe we tried to make it too user friendly with that comment in the file. Maybe that comment in the file should have been in the docs for people who want to know what this file is all about.

cducrest: `Then when the server crashes with cannot write file whatever -> the user can just create and doesn't have to deal with anything else.`
That's what this PR does, isn't it? The id can be added as first thing, or second thing regardless

compojoom: What I'm trying to say that the whole idea of having the first line as a comment seems stupid now. Because of that comment you now write code that checks whether the first line is a comment or not. Why do we have to add more code, if we can just say - fuck that. This file contains only an id...
Than there is no reason for the existence of your code and we have one less thing that test-coverage will complain about.

compojoom (approved):

PR #577, opened 2021-03-18
No longer push event information for balance updates from filter

cducrest:
Use ethindex updates instead.
The updates were incorrect because they were using information taken
from the graph before it was updated (the filter being triggered
faster). Now react on graph feed to publish events about trustline
updates and balance updates


compojoom (approved):
nice!

PR #590, opened 2021-05-03
Lower indexer required confirmation in tests to 50

cducrest:
This speeds up tests using the indexer a lot.
This should also make them non-flaky as the tests won't have to
wait for as long for the indexer to sync.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,6 +9,8 @@ executors:
   ubuntu-builder:
     docker:
       - image: trustlines/builder:master61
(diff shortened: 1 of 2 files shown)


compojoom (approved):

compojoom:
does the medium class actually bring us anything?

cducrest:
It's the default one, I just made it explicit.
The bigger class made the test like 2-3% faster which could just be random, however they did help me to debug how to make the test faster and non-flaky because they failed in a more stable way.

Issue #580, opened 2021-03-25
Relay server crashes with LoopExit


cducrest:
When Sascha started the version 0.20.2 of the relay, it crashed with traceback:
```
2021-03-25 09:48:38,193 INFO     [main] trustlines: Starting relay server version 0.20.2
2021-03-25 09:48:38,217 INFO     [main] web3provider: Autodetect provider from uri ws://ws.ethereum.trustlines.tlbc.parity.trustlines-relay-0001
2021-03-25 09:48:38,219 INFO     [main] web3provider: Autodetected WebsocketProvider
2021-03-25 09:48:38,221 WARNING  [main] relay: No account configured
2021-03-25 09:48:38,221 INFO     [main] node: Assuming connected to parity node: Enabling parity-only rpc methods.
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/opt/relay/lib/python3.8/site-packages/sentry_sdk/integrations/threading.py", line 69, in run
    reraise(*_capture_exception())
  File "/opt/relay/lib/python3.8/site-packages/sentry_sdk/_compat.py", line 57, in reraise
    raise value
  File "/opt/relay/lib/python3.8/site-packages/sentry_sdk/integrations/threading.py", line 67, in run
    return old_run_func(self, *a, **kw)
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/relay/lib/python3.8/site-packages/web3/providers/websocket.py", line 38, in _start_event_loop
    loop.run_forever()
  File "/usr/lib/python3.8/asyncio/base_events.py", line 570, in run_forever
    self._run_once()
  File "/usr/lib/python3.8/asyncio/base_events.py", line 1823, in _run_once
    event_list = self._selector.select(timeout)
  File "/opt/relay/lib/python3.8/site-packages/gevent/selectors.py", line 201, in select
    self._ready.wait(timeout)
  File "src/gevent/event.py", line 163, in gevent._gevent_cevent.Event.wait
  File "src/gevent/_abstract_linkable.py", line 521, in gevent._gevent_c_abstract_linkable.AbstractLinkable._wait
  File "src/gevent/_abstract_linkable.py", line 487, in gevent._gevent_c_abstract_linkable.AbstractLinkable._wait_core
[...]

PR #592, opened 2021-05-04
Drop support for websockets

cducrest:
It was broken due to conflict in between asyncio and gevent

closes https://github.com/trustlines-protocol/relay/issues/580


compojoom (approved):
Did we update the docs?

cducrest:
I think I updated the only place I found infos about websocket in the docs here: https://github.com/trustlines-protocol/tl-dev-docs/pull/48
Did I miss something?

PR #594, opened 2021-05-10
Update docker-compose setup for using tlbc and updated config

cducrest:


compojoom (approved):

compojoom (changes requested):
Can you actually change the relay-config.toml to `config.dist.toml`? Having the .dist ext in the name makes it clear that one needs to rename this file:

https://github.com/trustlines-protocol/tl-dev-docs/pull/50

compojoom (approved):

